PR #6, opened 2022-09-23
add .status and .ok to response from fetch

karthik2804:


crates/spin-js-engine/sdk.js
```diff
@@ -10,6 +10,73 @@ function fetch(uri, options) {
         headers: {
             entries: () => Object.entries(headers)
         },
-        arrayBuffer: () => Promise.resolve(body)
+        arrayBuffer: () => Promise.resolve(body),
+        ok: (status > 199 && status < 300),
+        statusText: statusTexts[status]
```
dicej: I assume this should read `statusText: statusTextList[status]`, given the name you used below, right?

dicej (changes requested):
Looks good, but note the `statusTexts` vs. `statusTextList` name mismatch.

crates/spin-js-engine/sdk.js
karthik2804: Yes, Good catch. I missed it while renaming things. Fixed it now.

dicej (approved):

PR #7, opened 2022-09-23
add text and json methods to fetch

karthik2804:


dicej (approved):
Looks good.  Note that currently we don't include `TextDecoder` as part of sdk.js, so we'll need to add that in by making sdk.js an NPM project that uses e.g. https://www.npmjs.com/package/fast-text-encoding.

PR #8, opened 2022-09-27
Add text-encoding as a part of sdk

### LINKED ISSUE #5: `TextEncoder` and `TextDecoder`
I'm trying to use an external library that needs a `TextDecoder`, and I had the imports in the following order:

```
import <some library that needs TextEncoder and TextDecoder>;
import { TextEncoder, TextDecoder } from "fast-text-encoding";

```

Building this results in `TextDecoder not defined`:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Uncaught ReferenceError: 'TextDecoder' is not defined
    at <anonymous> (script.js:1645)
    at <anonymous> (script.js:1882)
    at <eval> (script.js:1885)
', crates/spin-js-engine/src/lib.rs:193:15
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: the `wizer.initialize` function trapped

Caused by:
    wasm trap: wasm `unreachable` instruction executed
    wasm backtrace:
        0: 0x188002 - <unknown>!__rust_start_panic
    note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable to may show more debugging information
```

Swapping the imports order works around this issue. However, it would be great if `TextEncoder`, `TextDecoder` and others would be available without having to 1) care about the import order, and 2) manually add it as a dependency.

As a side note, I am bumping into a similar issue for `new URL`, but so far haven't been able to import a package that works around the issue.

karthik2804:

closes #5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,14 @@ target/release/spinjs: target/wasm32-wasi/release/spin-js-engine.wasm
 	SPIN_JS_ENGINE_PATH=../../target/wasm32-wasi/release/spin_js_engine.wasm \
 	cargo build --release
 
-target/wasm32-wasi/release/spin-js-engine.wasm: crates/spin-js-engine/src/lib.rs
+target/wasm32-wasi/release/spin-js-engine.wasm: crates/spin-js-engine/sdk.js crates/spin-js-engine/src/lib.rs
 	cd crates/spin-js-engine && \
 	QUICKJS_WASM_SYS_WASI_SDK_PATH=$(WASI_SDK) \
 	cargo build --release --target=wasm32-wasi
 
+crates/spin-js-engine/sdk.js: crates/spin-js-engine/src/js_sdk/sdk.js
+	cd crates/spin-js-engine/src/js_sdk && npx webpack --mode=production
+
 .PHONY: clean
 clean:
-	rm -rf dist target
+	rm -rf dist target crates/spin-js-engine/sdk.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
 /node_modules
 /dist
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -12,11 +12,14 @@ target/release/spinjs: target/wasm32-wasi/release/spin-js-engine.wasm
 	SPIN_JS_ENGINE_PATH=../../target/wasm32-wasi/release/spin_js_engine.wasm \
 	cargo build --release
 
-target/wasm32-wasi/release/spin-js-engine.wasm: crates/spin-js-engine/src/lib.rs
+target/wasm32-wasi/release/spin-js-engine.wasm: target/sdk.js crates/spin-js-engine/src/lib.rs
 	cd crates/spin-js-engine && \
 	QUICKJS_WASM_SYS_WASI_SDK_PATH=$(WASI_SDK) \
 	cargo build --release --target=wasm32-wasi
 
+target/sdk.js: crates/spin-js-engine/src/js_sdk/sdk.js
```
dicej: This refers to "target/sdk.js", but the command doesn't actually create a file at that location.  Did you mean "crates/spin-js-engine/sdk.js" here and above?

dicej (changes requested):
Looks great overall, but please see my note about the Makefile changes.

Makefile
karthik2804: Yes, I will make the change. I do not have too much experience writing makefiles.

dicej:
@karthik2804 Thanks for making that change.  One more thing: would you please add `crates/spin-js-engine/sdk.js` to the `clean` target in the makefile?

dicej (approved):

PR #10, opened 2022-09-28
Add more functions to fetch and url for sdk

karthik2804:

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -33,7 +33,7 @@ struct HttpRequest {
     method: String,
     uri: String,
     #[serde(default)]
-    headers: HashMap<String, String>,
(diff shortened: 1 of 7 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -103,10 +103,10 @@ fn spin_send_http_request(context: &Context, _this: &Value, args: &[Value]) -> R
                 .uri(request.uri.deref());
 
             if let Some(headers) = builder.headers_mut() {
-                for (key, value) in &request.headers {
+                for header in &request.headers {
```
dicej: What's the reason for this change?

karthik2804: This was an oversight on my part. It is unnecessary. I just rewrote that piece of code.

dicej (approved):

PR #11, opened 2022-09-29
add file read for static assets support

karthik2804:

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -1,5 +1,6 @@
 #![deny(warnings)]
 
+use std::path::PathBuf;
 use {
     anyhow::{anyhow, Result},
     http::{header::HeaderName, request, HeaderValue},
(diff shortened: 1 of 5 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -141,6 +143,21 @@ fn spin_send_http_request(context: &Context, _this: &Value, args: &[Value]) -> R
...
+    match args {
+        [filename] => {
+            let  deserializer = &mut Deserializer::from(filename.clone());
+            let filename = PathBuf::deserialize(deserializer)?;
+            let buffer = fs::read(filename)?;
+            let mut serializer = Serializer::from_context(context)?;
+            buffer.serialize(&mut serializer)?;
+            Ok(serializer.value)
+        }
+        _ => Err(anyhow!("Could not read file")),
```
dicej: I'd suggest changing the error message here to something like "Expected a single file name argument to `read_file`, but got {} arguments"

dicej (changes requested):
Looks good, but please see the note about the `read_file` error message.

dicej (approved):

PR #12, opened 2022-09-29
Add examples

karthik2804:


dicej (commented):
This looks awesome!

Just one suggestion: I think we should get rid of the Makefiles completely.  Each example shouldn't be responsible for building `spin-js-engine` and `spin-js-cli`.  In other words, they should just assume `spinjs` is in the user's PATH and use it.  And since not all JS developers will have `make`, we should a more portable NPM build script to run `webpack` and `spinjs`.

Eventually, I'd like to publish `spinjs` as an NPM package if possible (e.g. https://stackoverflow.com/questions/67766855/distribute-a-binary-executable-with-a-nodejs-module-when-publishing-to-npmjs-com) so any NPM-based JS project can just declare it as a dependency.

In the meantime, let's keep the examples simple and portable so they don't require having `make` or `cargo` installed.

dicej (approved):

PR #13, opened 2022-10-03
Add examples for typescript and add starter templates

karthik2804:


examples/typescript/planetscale/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const config = {
+   host: '<host>',
+   username: '<username>',
+   password: '<password>'
+}
+
+const encoder = new TextEncoder()
+
+export async function handleRequest(_request: any) {
```
radu-matei: Is this the first step here?
Ideally, this would be `request: HttpRequest`, and we would import the actual types.

karthik2804: I do not have any experience with typescript. I will look into it.

examples/typescript/outbound-http/src/SpinSdkTypes.ts
```diff
@@ -0,0 +1,16 @@
+interface HttpRequest {
```
radu-matei: These are types that we will want to extract in a package and publish to NPM.

dicej:
@karthik2804 Let's pair on this when you're back from KubeCon.  I think we have a bit more work to do here before we can call TypeScript support "done".

templates/http-ts/content/src/SpinSDK.d.ts
```diff
@@ -0,0 +1,14 @@
+export declare interface HttpRequest {
```
dicej: Can we remove this file and use the one generated from sdk.ts instead?

templates/http-ts/content/types/lib/index.d.ts
```diff
@@ -0,0 +1,16 @@
+interface HttpRequest {
```
dicej: As above: can we remove this file?

types/lib/index.d.ts
```diff
@@ -0,0 +1,43 @@
+interface SpinConfig {
```
dicej, line 1: Can we remove this?

dicej (commented):
Looks great overall.  Can we remove the redundant type declaration files and only use the one generated from sdk.ts?

types/lib/index.d.ts
karthik2804: This file is just a copy of the generated from sdk.ts so as to package it as an npm package.

templates/http-ts/content/types/lib/index.d.ts
karthik2804: Done

templates/http-ts/content/src/SpinSDK.d.ts
karthik2804: Done

karthik2804:
Any ideas on how we can keep the index file in the `types` folder in sync with the source of the SDK? 

- Potentially add CI to check if generated `d.ts` is the same as the one in the `types` folder.
- Should we move the `types` folder into the crates as well so that we can directly generate the types into the npm package? (Wondering how deployments would look)

dicej (approved):
Let's go with this as-is for now, and we can work on de-duplicating the type declarations later.

Thanks for working on this, @karthik2804 !

PR #21, opened 2022-11-03
refactor SDK and add types to URL

karthik2804:


crates/spin-js-engine/src/js_sdk/modules/statusTextList.ts
```diff
@@ -0,0 +1,65 @@
+/** @internal */
+export const statusTextList: { [key: string]: string } = {
```
dicej, line 2: Should this be: `[key: number]` instead?

dicej (commented):
Looks great -- just one question above.

crates/spin-js-engine/src/js_sdk/modules/statusTextList.ts
karthik2804: I believe that all the keys in javascript are strings. To make it clearer, I think I can wrap all the keys with `"` if that would make sense.

dicej: No worries -- I guess if TypeScript knows what you mean, it's okay.

dicej (approved):

PR #22, opened 2022-11-04
Add fsPromise.readdir with minimal implementation for file types

karthik2804:
Implemented a minimal read directory based on `nodejs`. I also added the option to request for filetypes. It does not have all the properties provided by the node but supports the major ones. 

While implementing the read directory I was wondering if we should implement `glob` as well as that seems to be straightforward to implement and might be useful.

### DIFF
--- a/crates/spin-js-engine/src/js_sdk/modules/fsPromises.ts
+++ b/crates/spin-js-engine/src/js_sdk/modules/fsPromises.ts
@@ -1,19 +1,47 @@
+interface ReadDirOptions {
+    withFileTypes: boolean
+}
+
+interface ReadDirWithFileTypes {
+    name: string
+    isFile: () => boolean
+    isDirectory: () => boolean
+    isSymboliclink: () => boolean
+}
+
 declare const _fsPromises: {
     readFile: (arg0: string) => ArrayBuffer
+    readDir: {
+        (arg0: string): Array<string>
+        (arg0: string, arg1: ReadDirOptions & object): Array<ReadDirWithFileTypes>
+    }
 }
 
 /** @internal */
 const fsPromises = {
-    readFile: (filename: string) =>  {
+    readFile: (filename: string) => {
         return Promise.resolve(_fsPromises.readFile(filename))
+    },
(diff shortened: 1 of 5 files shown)


crates/spin-js-engine/src/js_sdk/modules/fsPromises.ts
```diff
@@ -1,19 +1,54 @@
...
 /** @internal */
 const fsPromises = {
-    readFile: (filename: string) =>  {
+    readFile: (filename: string) => {
         return Promise.resolve(_fsPromises.readFile(filename))
+    },
+    readdir: (dirname: string, options?: ReadDirOptions) => {
+        if (options && options.withFileTypes == true) {
+            let res = _fsPromises.readDir(dirname, options)
+            // let ret: Array<ReadDirWithFileTypes> = []
```
dicej: Can we just remove this commented-out code?

crates/spin-js-engine/src/lib.rs
```diff
@@ -143,19 +145,97 @@ fn spin_send_http_request(context: &Context, _this: &Value, args: &[Value]) -> R
...
+                    context.wrap_callback({
+                        let e = e.clone();
+                        move |context, _, _| context.value_from_bool(e.file_type()?.is_dir())
+                    })?,
+                )?;
+
+                entry.set_property(
+                    "isSymbolicLink",
+                    context.wrap_callback({
+                        let e = e.clone();
```
dicej: Can you run `cargo clippy` on this code?  I think `clippy` will tell you this clone is unnecessary.

karthik2804: Will check

karthik2804: `clippy` does not say anything about those clones. It did find a couple of other unnecessary stuff that I have fixed.

crates/spin-js-engine/src/js_sdk/modules/fsPromises.ts
karthik2804: done

dicej (approved):

PR #27, opened 2022-11-09
feat/add_crypto.getRandomValues

karthik2804:


crates/spin-js-engine/src/lib.rs
```diff
@@ -254,8 +254,12 @@ fn get_glob(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
     }
 }
 
+fn get_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value> {
+    return context.value_from_u32(thread_rng().gen_range(0..=255));
+}
+
 fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value> {
-    return context.value_from_f64(thread_rng().gen_range(0.0_f64..1.0));
+    return context.value_from_f64(thread_rng().gen_range(0.0_f64..=1.0));
```
dicej: I don't think this is correct -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random says the range should exclude 1.0, not include it.

karthik2804: Ah good catch! missed that.

karthik2804: Fixed it.

dicej (approved):

PR #32, opened 2022-11-16
eliminate `unsafe` blocks in `spin-js-engine`

dicej:
Since Wasm is single-threaded (for now), there was nothing unsound about our use of `unsafe`, but it was annoying and added the extra maintenance burden of making sure it stayed sound going forward.  With help from `send_wrapper` and `once_cell`, we can get rid of the `unsafe` blocks at the cost of slightly more verbose type declarations.

This also fixes a few clippy warnings.


karthik2804 (approved):
LGTM!

PR #33, opened 2022-11-17
switch back to upstream Shopify/javy repo

dicej:
All our patches have been merged upstream, so we no longer need to use the fork. I've also run `cargo update` to update all the Rust dependencies.


karthik2804 (approved):
LGTM!

PR #36, opened 2022-11-19
feat/add_redis

karthik2804:

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -12,7 +12,7 @@ use {
     spin_sdk::{
         config,
         http::{Request, Response},
-        http_component, outbound_http,
+        http_component, outbound_http, redis,
     },
     std::{
         collections::HashMap,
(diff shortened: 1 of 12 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -260,6 +260,107 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
     context.value_from_f64(thread_rng().gen_range(0.0_f64..1.0))
 }
 
+fn redis_get(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
+    match args {
+        [address, key] => {
+            let addr_deserializer = &mut Deserializer::from(address.clone());
```
dicej: We should probably implement `TryFrom<&Value>` for `String` and move the deserialization code into there.  That way, we can simplify all these conversions to e.g. `let address = address.try_into()?`.

crates/spin-js-engine/src/lib.rs
```diff
@@ -260,6 +260,107 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+    }
+}
+fn redis_incr(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
+    match args {
+        [address, key] => {
+            let addr_deserializer = &mut Deserializer::from(address.clone());
+            let address = &String::deserialize(addr_deserializer)?;
+            let key_deserializer = &mut Deserializer::from(key.clone());
+            let key = &String::deserialize(key_deserializer)?;
+            let buffer = redis::incr(address, key).map_err(|_| anyhow!("Error executing Redis incr command"))?;
```
dicej: I'd recommend renaming `buffer` to `value` since it's just an integer, not a buffer.

crates/spin-js-engine/src/lib.rs
```diff
@@ -260,6 +260,107 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+        [address, key] => {
+            let addr_deserializer = &mut Deserializer::from(address.clone());
+            let address = &String::deserialize(addr_deserializer)?;
+            let key_deserializer = &mut Deserializer::from(key.clone());
+            let key = Vec::<String>::deserialize(key_deserializer)?;
+            let mut keys  = Vec::new();
+            for i in &key {
+                keys.push(i.as_str());
+            }
+            let buffer = redis::del(address, keys.as_slice()).map_err(|_| anyhow!("Error executing Redis del command"))?;
```
dicej: As above, I recommend calling this `value` rather than `buffer`.

crates/spin-js-engine/src/lib.rs
```diff
@@ -260,6 +260,107 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+    match args {
+        [address, key, value] => {
+            let addr_deserializer = &mut Deserializer::from(address.clone());
+            let address = &String::deserialize(addr_deserializer)?;
+            let key_deserializer = &mut Deserializer::from(key.clone());
+            let key = &String::deserialize(key_deserializer)?;
+            let value_deserializer = &mut Deserializer::from(value.clone());
+            let value = ByteBuf::deserialize(value_deserializer)?;
+            redis::set(address, key, &value).map_err(|_| anyhow!("Error executing Redis set command"))?;
+            let mut serializer = Serializer::from_context(context)?;
```
dicej: I'd just return `context.undefined_value()` here instead of doing the whole serialization thing.

crates/spin-js-engine/src/lib.rs
```diff
@@ -260,6 +260,107 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+    match args {
+        [address, key, value] => {
+            let addr_deserializer = &mut Deserializer::from(address.clone());
+            let address = &String::deserialize(addr_deserializer)?;
+            let channel_deserializer = &mut Deserializer::from(key.clone());
+            let channel = &String::deserialize(channel_deserializer)?;
+            let value_deserializer = &mut Deserializer::from(value.clone());
+            let value = ByteBuf::deserialize(value_deserializer)?;
+            redis::publish(address, channel, &value).map_err(|_| anyhow!("Error executing Redis publish command"))?;
+            let mut serializer = Serializer::from_context(context)?;
```
dicej: As above: I suggest using `context.undefined_value()`.

types/lib/modules/spinSdk.d.ts
```diff
@@ -19,6 +19,13 @@ interface HttpResponse {
 declare type HandleRequest = (request: HttpRequest) => Promise<HttpResponse>;
 interface SpinSDK {
     config: SpinConfig;
+    redis: {
+        del: (address: string, keys: Array<string>) => {};
```
dicej: Can we return equivalent types to outbound-redis.wit here?  e.g. `BigInt` for `del` and `incr` and `undefined` (i.e. nothing) for `publish` and `set`?

karthik2804:
Implemented all the changes except the `TryFrom`.

dicej (approved):

PR #38, opened 2022-11-22
switch from vector to hashmap for headers in handleRequest

karthik2804:

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -41,6 +41,16 @@ struct HttpRequest {
     body: Option<ByteBuf>,
 }
 
+// Seperate Structure to add headers without having to use serde
+// serde causes issues with headers as it sanitzes keys leading to headers being invalid
+// (eg) content-type becomes contentType.
+#[derive(Serialize, Deserialize, Debug)]
+struct JSHttpRequest {
+    method: String,
+    uri: String,
+    body: Option<ByteBuf>,
+}
+
 #[derive(Serialize, Deserialize, Debug)]
 struct HttpResponse {
     status: u16,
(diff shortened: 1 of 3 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -447,20 +457,18 @@ fn handle(request: Request) -> Result<Response> {
     process.set_property("env", env)?;
 
     global.set_property("process", process)?;
+    
+    let headers = context.object_value()?;
+    for (key, value) in request.headers().iter() {
+        let header_value =  context.value_from_str(&str::from_utf8(value.as_bytes()).unwrap().to_owned());
```
dicej: I think we should probably just let the error bubble up here (which will avoid the need for the `if let` below).
```suggestion
        let header_value =  context.value_from_str(&str::from_utf8(value.as_bytes()).unwrap().to_owned())?;
```

karthik2804: fixed it. Removed the unwrap as well.

crates/spin-js-engine/src/lib.rs
```diff
@@ -447,20 +457,16 @@ fn handle(request: Request) -> Result<Response> {
     process.set_property("env", env)?;
 
     global.set_property("process", process)?;
+    
+    let headers = context.object_value()?;
+    for (key, value) in request.headers().iter() {
+        let header_value =  context.value_from_str(&str::from_utf8(value.as_bytes()).to_owned()?)?;
```
dicej:
```suggestion
        let header_value =  context.value_from_str(str::from_utf8(value.as_bytes())?)?;
```

karthik2804: fixed

dicej (approved):

Issue #41, opened 2022-11-25
HTTP header names are not parsed correct on responses


mikkelhegn:
Headers with '-' becomes formatted as an underscore '_' i.e., `content-type` become `content_type`

<img width="1399" alt="image" src="https://user-images.githubusercontent.com/12242001/204009971-1d3f45b1-520a-4fba-bb32-afd1fcb11ec7.png">

karthik2804:
closed with #42

PR #42, opened 2022-11-28
use case-preserving `Serializer` and `Deserializer`

dicej:
This addresses https://github.com/fermyon/spin-js-sdk/issues/41


karthik2804 (approved):
LGTM!

PR #47, opened 2022-11-30
remove binaryen for windows

karthik2804:

### DIFF
--- a/crates/spin-js-cli/src/main.rs
+++ b/crates/spin-js-cli/src/main.rs
@@ -1,8 +1,9 @@
 #![deny(warnings)]
 
+#[cfg(not(target_os = "windows"))]
+use binaryen::{CodegenConfig, Module};
 use {
     anyhow::{bail, Context, Result},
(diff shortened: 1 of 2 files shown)


crates/spin-js-cli/src/main.rs
```diff
@@ -31,28 +32,39 @@ fn main() -> Result<()> {
...
-
-        if let Ok(mut module) = Module::read(&wasm) {
-            module.optimize(&codegen_cfg);
-            module
-                .run_optimization_passes(vec!["strip"], &codegen_cfg)
-                .unwrap();
-            wasm = module.write();
-        } else {
-            bail!("Unable to read wasm binary for wasm-opt optimizations");
+        #[cfg(target_os = "windows")]
```
dicej, line 36: Could you add a comment here explaining why we don't use `binaryen` on Windows?

karthik2804: Done.

dicej (approved):

PR #60, opened 2022-12-06
update examples, templates and readme

karthik2804:


examples/javascript/outbound-http/spin.toml
```diff
@@ -15,6 +12,5 @@ files = ["src/index.html"]
 [component.trigger]
 route = "/hello"
 [component.config]
```
dicej: We should probably remove this line, too, if there are no config values.

dicej (approved):

PR #65, opened 2022-12-08
switch back to upstream Javy

dicej:
Now that https://github.com/Shopify/javy/pull/150 has been merged, we can use the upstream repo again, and no longer need to explicitly disable case conversion, since `quickjs-wasm-rs` no longer does case conversion at all.


karthik2804 (approved):
LGTM!

PR #66, opened 2022-12-10
Add snippets to enable `spin add`

radu-matei:
cc @itowlson 

Signed-off-by: Radu Matei <radu.matei@fermyon.com>

### DIFF
--- a/templates/http-js/metadata/snippets/component.txt
+++ b/templates/http-js/metadata/snippets/component.txt
@@ -1,0 +1,9 @@
+[[component]]
+id = "{{project-name | kebab_case}}"
+source = "{{ output-path }}/target/spin-http-js.wasm"
+[component.trigger]
+route = "{{http-path}}"
+[component.build]
+command = "npm run build"
+workdir = "{{ output-path }}"
+
--- a/templates/http-js/metadata/spin-template.toml
+++ b/templates/http-js/metadata/spin-template.toml
@@ -2,6 +2,12 @@ manifest_version = "1"
 id = "http-js"
 description = "HTTP request handler using Javascript"
 
+[add_component]
+skip_files = ["spin.toml"]
+skip_parameters = ["http-base"]
+[add_component.snippets]
+component = "component.txt"
+
 [parameters]
 project-description = { type = "string",  prompt = "Project description", default = "" }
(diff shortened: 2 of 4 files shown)


karthik2804:
I run into an error when I try to add a TypeScript component but it succeeds when using a js-component.

```bash
# errors out
$ spin add http-ts test-component --accept-defaults
Error: liquid: Unknown variable
  with:
    requested variable=msg


# succeeds
$ spin add http-js test-component --accept-defaults
```

radu-matei:
Make sure to uninstall the TS template before, I just re-ran `spin add` with a TypeScript component and everything seems to work fine for me.

karthik2804:
That solved it! I had to install (update) the templates from your branch. This works as expected.
```bash
spin templates install --update --git https://github.com/radu-matei/spin-js-sdk/ --branch templates-add-component
```

I keep running into the same error when I try to install the templates from a local directory where the PR was checked out with the following command

```bash
spin templates install --update --dir ./
```

Not sure if this is an issue only on my end. I will take another look tomorrow.

itowlson (approved):
This looks fine in itself, and I'm happy for us to merge it.

But it would be good to modify the standard template to be consistent with the snippet, e.g. deriving `component.id` instead of fixing it to `hello`, deriving `component.trigger.route`.

Also, and I guess out of scope for this, for the standard template we should:

* Derive the application `name`
* Derive the application `trigger.base`

At the moment it looks like we prompt for these but don't actually use them, though I might be missing something - I've only read the `spin.toml` not tried running it.

I would also suggest that we consider building to something like `{{ project-name | kebab_case }}.wasm` rather than always `spin-http-js.wasm`, to make visual distinction easier if modules get copied around.  Not a big deal though, e.g. if the tooling that makes this onerous/impractical.

karthik2804 (approved):

karthik2804:
@radu-matei Do you want to include the fixes to the template to derive `name`, `trigger.base` and `path` as pointed above or should I make a separate PR for that.

karthik2804:
@radu-matei Do you want to include the fixes to the template to derive `name`, `trigger.base` and `path` as pointed above or should I make a separate PR for that.

radu-matei:
Feel free to create a separate PR, @karthik2804!
Thanks!

PR #68, opened 2022-12-15
fix templates to use liquid filters

karthik2804:


dicej (approved):

itowlson (commented):
It would be good also to update `package.json` to reflect the `project-name` - currently the package will always be called `hello_world`.

Otherwise LGTM - thanks!

PR #76, opened 2022-12-20
Add redis commands SADD SMEMBERS SREM

karthik2804:

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -305,17 +305,13 @@ fn redis_incr(context: &Context, _this: &Value, args: &[Value]) -> Result<Value>
     }
 }
 
-
 fn redis_del(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
     match args {
         [address, key] => {
             let address = &deserialize_helper(address)?;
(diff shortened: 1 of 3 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -348,6 +346,72 @@ fn redis_set(context: &Context, _this: &Value, args: &[Value]) -> Result<Value>
...
 }
 
+fn redis_sadd(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
+    match args {
+        [address, key, values] => {
+            let address = &deserialize_helper(address)?;
+            let key = &deserialize_helper(key)?;
+            let deserializer = &mut Deserializer::from(values.clone());
+            let value_array = Vec::<String>::deserialize(deserializer)?;
+            let mut values = Vec::new();
```
dicej: Nit: this could be written more succinctly and efficiently as `let values = value_array.iter().map(|s| s.as_str()).collect::<Vec<_>>();`.  Same goes for `redis_srem`.

dicej (approved):

crates/spin-js-engine/src/lib.rs
karthik2804: Fixed it. Thanks!

PR #81, opened 2022-12-21
remove workarounds for `quickjs-wasm-rs` case conversion

dicej:
Now that `quickjs-wasm-rs` no longer does object key case conversion when serializing and deserializing, there's no need for complicated workarounds.

### DIFF
--- a/crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
+++ b/crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
@@ -9,14 +9,10 @@ interface BaseHttpRequest {
     method: string
     uri: string
     body?: ArrayBuffer
-}
-
-interface SpinHttpRequest extends BaseHttpRequest {
-    headers: Array<[string, string]>
+    headers: Record<string, string>
 }
 
 interface HttpRequest extends BaseHttpRequest {
-    headers: Record<string, string>
     json:() => object
     text: () => string
 }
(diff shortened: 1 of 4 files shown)


dicej:
@karthik2804 Please let me know if there are any files that need to be re-generated based on these changes.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -49,7 +45,7 @@ interface SpinSDK {
 
 interface FetchOptions {
     method: string
-    headers: object
```
karthik2804, line 52: We need to add another field `body: string`

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -68,15 +64,10 @@ interface FetchResult {
...
 /** @internal */
 function fetch(uri: string, options?: FetchOptions) {
-    let reqHeaders: Array<[string, string]> = []
-    if (options && options.headers) {
-        reqHeaders = Object.entries(options.headers)
-    }
     const { status, headers, body } = spinSdk.http.send({
         method: (options && options.method) || "GET",
         uri,
-        ...(options || {}),
```
karthik2804, line 78: Removing this line causes, the ability to pass the body to the fetch request to be removed. see above for the changes suggested in this regard.

karthik2804 (approved):

karthik2804:
@dicej this one looks good to merge to me after the first commit is signed.

PR #92, opened 2022-12-29
fix URLSearchParams.entries and add iterator

karthik2804:
This commit changes the `entries` method to be consistent with the browser implementation.

### DIFF
--- a/crates/spin-js-engine/src/js_sdk/modules/url.ts
+++ b/crates/spin-js-engine/src/js_sdk/modules/url.ts
@@ -81,7 +81,25 @@ class URLSearchParams {
         delete this.queryParams[key]
     }
     entries() {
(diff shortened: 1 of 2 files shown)


crates/spin-js-engine/src/js_sdk/modules/url.ts
```diff
@@ -81,7 +81,27 @@ class URLSearchParams {
...
+        let iterLength = arr.length
+        let iterIndex = 0
+        return {
+            next: function() {
+                return iterIndex < iterLength ?
+               {value: arr[iterIndex++], done: false} :
+               {done: true};
+            }
+        }
+        return arr.entries()
```
dicej:
```suggestion
```

dicej (approved):

PR #35, opened 2022-11-18
Readme and example nits

mikkelhegn:
Add CMake as a dependency
Updated build command in javascript/hello-world example

Signed-off-by: Mikkel Mørk Hegnhøj <mikkel@fermyon.com>

PR #97, opened 2023-01-06
add test action

karthik2804:
This PR adds some tests in CI. This PR also adds some extra methods to fetch headers (`get` and `has`) as per https://developer.mozilla.org/en-US/docs/Web/API/Headers.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: "Install Wasm Rust target"
+        shell: bash
+        run: rustup target add wasm32-wasi
+
+      - name: Install nodejs
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Setup cmake
```
dicej, line 44: Why is CMake needed?

dicej (approved):

.github/workflows/test.yaml
karthik2804: It is a needed package for the build.
https://github.com/fermyon/spin-js-sdk/blob/main/README.md#L84

https://github.com/fermyon/spin-js-sdk/pull/35

dicej: Ah yeah, I forgot about that.  Thanks.

PR #99, opened 2023-01-06
add router to the SDK

karthik2804:
Adds a router to the SDK using the `itty-router`

Usage example of the router 

```js
import { HandleRequest, HttpRequest, HttpResponse } from "spin-sdk"

const encoder = new TextEncoder()

router.get("/", () => {return {status: 200, body: encoder.encode("default route").buffer}})
router.get("/test", () => {return {status: 200, body: encoder.encode("test route").buffer}})

export const handleRequest: HandleRequest = async function (request: HttpRequest): Promise<HttpResponse> {
    
    return await router.handle(request)
}
```

### DIFF
--- a/crates/spin-js-engine/src/js_sdk/modules/router.ts
+++ b/crates/spin-js-engine/src/js_sdk/modules/router.ts
@@ -1,0 +1,85 @@
+/** @internal */
+import { Router as _router } from 'itty-router'
+import { HttpRequest } from './spinSdk'
+
+declare type GenericTraps = {
+    [key: string]: any;
+};
+
+export declare type RequestLike = {
+    method: string;
+    url: string;
+} & GenericTraps;
+
+declare type IRequest = {
+    method: string;
+    url: string;
+    params: {
+        [key: string]: string;
+    };
+    query: {
+        [key: string]: string | string[] | undefined;
+    };
+    proxy?: any;
+} & GenericTraps;
+
+interface RouteHandler {
+    (request: IRequest, ...args: any): any;
+}
+
+declare type RouteEntry = [string, RegExp, RouteHandler[]];
+declare type Route = <T extends RouterType>(path: string, ...handlers: RouteHandler[]) => T;
+declare type RouterHints = {
+    all: Route;
+    delete: Route;
+    get: Route;
+    options: Route;
+    patch: Route;
+    post: Route;
+    put: Route;
+};
+declare type RouterType = {
+    __proto__: RouterType;
+    routes: RouteEntry[];
+    handle: (request: RequestLike, ...extra: any) => Promise<any>;
+} & RouterHints;
+
+interface routerType {
+    all(path: string, ...handlers: RouteHandler[]): RouterType 
+    delete(path: string, ...handlers: RouteHandler[]): RouterType 
+    get(path: string, ...handlers: RouteHandler[]): RouterType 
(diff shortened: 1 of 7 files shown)


dicej (commented):
I have a deep-seated aversion to mutable global state, so the global router singleton feels very awkward to me.  Also, Itty seems like a fairly popular package which many JS/TS devs may already be familiar with, so I feel like we shouldn't hide the fact that we're using it.

Could we just include it as a built-in package and let app devs use like they usually would, e.g. `import { Router } from 'itty-router'`?  It's already a TypeScript-first library, so we wouldn't need to add any explicit type declarations -- just use the ones that came with the library.

I feel like the existing Itty API is natural and straightforward, so I don't think we need to restrict or simplify it.

karthik2804:
The intention behind it is to simplify the handle function. If users import it manually into their app

```js
await router.handle({
    method: request.method,
    url: spin.headers["spin-full-url"]
})
```

While in the case that it is a part of the SDK as per the conversation with @radu-matei, it simplifies down to 
```js
await router.handle(request)
```

Don't have a really strong opinion on either of the approaches in particular. thoughts?

dicej:
Thanks for explaining, @karthik2804.  I would vote for either adding a `handleRequest` method to `Router` that takes an `HttpRequest` (e.g. using a `Proxy`) or adding a top-level `spinSdk` function `handleRequest` function that takes both a `Router` and an `HttpRequest`.

karthik2804:
@dicej As long as we include the router in the SDK, we would have to inline the types for the router in the SDK file otherwise, it will introduce a dependency on the `itty-package` for the types package which is fine I guess.

dicej:
Is there anything wrong with adding the `itty-package` dependency to the `types` package?  Are we trying to avoid depending on Itty directly so that we have an option to replace it later?  If so, that's fine, just trying to understand the constraints here.

I'm fine with abstracting away the Itty dependency if desired, but I'd still like to avoid the global singleton.  Perhaps something like:

```js
import { Router, HttpRequest } from "@fermyon/spin-sdk"
const router = Router()
...
router.handle(request)
```

karthik2804:
We would not be able to do the following 
```
import { Router, HttpRequest } from "@fermyon/spin-sdk"
```

as that would mean that the `@fermyon/spin-sdk` has to provide an implementation that with the way the project is structured can provide only type definitions for functions that will be available at runtime.

radu-matei:
My main ask here is that our handler can be:

```typescript
export const handleRequest: HandleRequest = async function (request: HttpRequest): Promise<HttpResponse>{    
    return await router.handle(request)
}
```

I think this is a really good experience (as opposed to manually passing the request objects to the router).

karthik2804:
The Router function is now exposed through utils and can now be used as 

```js
import { HandleRequest, HttpRequest, HttpResponse } from "spin-sdk"

const encoder = new TextEncoder()

let router = utils.Router()

router.get("/hello", () => console.log("reached hello"))

export const handleRequest: HandleRequest = async function (request: HttpRequest): Promise<HttpResponse> {

    await router.handleRequest(request)

    return {
        status: 200,
        headers: {"foo": "bar"},
        body: encoder.encode("Hello from JS-SDK").buffer
    }
}

```

dicej (approved):
My ideal API would avoid adding anything to the global namespace and instead make the `Router` type importable from the `spin-sdk` module, but I understand that achieving that is difficult given how everything fits together.

Issue #100, opened 2023-01-14
Add `crypto.createHmac` and `crypto.createHash`


radu-matei:
These functions are extremely useful for computing signatures for payloads.

karthik2804:
Would the `digest` method from subtleCrypto implementation on the browser work for the use cases of `createHash`? 

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest 

Or would the `update` and `digest` methods on the node API be preferred?

PR #101, opened 2023-01-25
add sha256 and sha512 digests

karthik2804:
implements both browser and nodej API for creating hashes using either sha256 and sha512 as per #100.


crates/spin-js-engine/src/js_sdk/modules/crypto.ts
```diff
@@ -17,5 +22,39 @@ export const crypto = {
...
+            }
+        }
+    },
+    createHash: function (algorithm: string) {
+        let data: ArrayBuffer
+        return {
+            update(content: string | ArrayBuffer, inputEncoding: string = "string") {
+                if (inputEncoding != "string") {
+                    // @ts-ignore
+                    data = content
```
dicej: My understanding was that data could be streamed to the digest using `update`, meaning `update` should append to, not replace, the internal buffer.  The docs are ambiguous about this, though: https://nodejs.org/api/crypto.html#cipherupdatedata-inputencoding-outputencoding.

crates/spin-js-engine/src/lib.rs
```diff
@@ -256,6 +257,30 @@ fn get_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+            let algorithm = &deserialize_helper(algorithm)?;
+            let content_deserializer = &mut Deserializer::from(content.clone());
+            let content = ByteBuf::deserialize(content_deserializer)?;
+            let value: [u8; 32];
+            if algorithm == "sha256" {
+                let hashvalue = sha2::Sha256::digest(content);
+                value = hashvalue.as_slice().try_into().expect("Wrong length");
+            } else if algorithm == "sha512" {
+                let hashvalue = sha2::Sha512::digest(content);
+                value = hashvalue.as_slice().try_into().expect("Wrong length");
```
dicej: SHA-512 generates a 64 byte hash, not a 32 byte one, right?  Can we just leave them as `Vec<u8>` instead of converting them to arrays?

crates/spin-js-engine/src/js_sdk/modules/crypto.ts
karthik2804: Good point! fixed it. I was a little confused by the docs but have fixed it now.

crates/spin-js-engine/src/lib.rs
karthik2804: Fixed it.

crates/spin-js-engine/src/js_sdk/modules/crypto.ts
```diff
@@ -17,5 +22,41 @@ export const crypto = {
...
+    },
+    createHash: function (algorithm: string) {
+        let data = new Uint8Array()
+        return {
+            update(content: string | Uint8Array, inputEncoding: string = "string") {
+                if (inputEncoding == "string") {
+                    // @ts-ignore
+                    content = encoder.encode(content)
+                }
+                let  mergedArray = new Uint8Array(data.length + content.length);
```
dicej, line 45: This looks correct, but it will be pretty inefficient for many calls to `update` (e.g. streaming a large download and hashing it incrementally).  Ideally, we'd expose three Rust functions to JS: `new_hasher`, `update_hasher`, and `finish_hasher` so we don't have to do so many allocations in JS.

Anyway, it's fine for now, but we'll want to revisit if performance becomes a problem.

dicej (approved):

PR #105, opened 2023-01-30
enable `Wizer::wasm_bulk_memory`

dicej:
As of Rust 1.67, the `wasm32-wasi` target seems to be using bulk memory operations by default, which means we need Wizer to support them, too.


karthik2804 (approved):
LGTM!

PR #115, opened 2023-02-09
make return type flexible

karthik2804:
This PR allows the return type to be more flexible on the body parameter

```js
export async function handleRequest(request) {

    return {
        status: 200,
        headers: { "foo": "bar" },
        body: "Hello from JS-SDK"
    }
}
```

This PR also adds a new handler function signature `handler(req, res)` which allows creating responses using chaining

```js
import { EventHandler, HttpRequest, HttpResponse } from "spin-sdk"

export const eventHandler: EventHandler = async function (request: HttpRequest, response: ResponseObject): Promise<void> {

    response.status(200).header("content-type", "text/html").body("Chaining works")
}
```

### DIFF
--- a/crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
+++ b/crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
@@ -3,7 +3,7 @@ require('fast-text-encoding')
 let encoder = new TextEncoder()
 
 /** @internal */
-import {statusTextList} from "./statusTextList"
+import { statusTextList } from "./statusTextList"
 
 interface SpinConfig {
     get(arg0: string): string
@@ -17,7 +17,7 @@ interface BaseHttpRequest {
 }
 
 interface HttpRequest extends BaseHttpRequest {
-    json:() => object
+    json: () => object
     text: () => string
 }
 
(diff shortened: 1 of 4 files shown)


crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -70,17 +70,20 @@ interface FetchResult {
...
-    let encodedBodyData
-    if (options && options.body) {
-        if(typeof (options.body) == "string"){
-            encodedBodyData = encoder.encode(options.body).buffer
-        } else if (ArrayBuffer.isView(options.body)) {
-            encodedBodyData = options.body.buffer
-        } else {
-            encodedBodyData = options.body
-        }
+function encodebody(body: ArrayBuffer | Uint8Array | string) {
```
radu-matei: nit: `encodeBody`

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -105,12 +108,76 @@ function fetch(uri: string, options?: FetchOptions) {
     })
 }
 
+class ResponseObject {
```
radu-matei: Would `ResponseBuilder` be a better name for this?

radu-matei (commented):
This looks great, the comment I have is that the two changes are logically separate, and I think they should have separate PRs.

karthik2804:
Will move the implementation of the alternate function signature to a separate PR.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
karthik2804: Will address in the subsequent PR for signature type

crates/spin-js-engine/src/lib.rs
```diff
@@ -507,10 +507,20 @@ fn do_init() -> Result<()> {
 
     let global = context.global_object()?;
 
-    let entrypoint = global.get_property("spin")?.get_property("handleRequest")?;
-
-    if !entrypoint.is_function() {
-        panic!("expected function named \"handleRequest\" in \"spin\"");
+    let entrypoint;
+    match global
```
dicej, line 511: I'd just use an `if` statement here -- no need to use `match` on a `bool` value.

dicej (approved):

PR #120, opened 2023-02-13
add crypto.verify and timingSafeEqual

### LINKED ISSUE #118: Implement `crypto.timingSafeEqual` and `crypto.subtle.verify`
This would improve the security when verifying digests and signatures, and it should use a constant-time algorithm.

See https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b

karthik2804:
This PR implements a minimal subset of the `crypto.verify`/`crypto.sublte.verify` API with support only for `HMAC-SHA256` and also adds support for `timingSafeEqual`

closes #118 


crates/spin-js-engine/src/lib.rs
```diff
@@ -497,6 +498,25 @@ fn set_timeout(context: &Context, _this: &Value, args: &[Value]) -> Result<Value
...
 }
 
+fn timing_safe_equals(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
+    match args {
+        [value1, value2] => {
+            let value1_deserializer = &mut Deserializer::from(value1.clone());
+            let value1 = ByteBuf::deserialize(value1_deserializer)?;
+            let value2_deserializer = &mut Deserializer::from(value2.clone());
+            let value2 = ByteBuf::deserialize(value2_deserializer)?;
+            match bool::from(value1.ct_eq(&value2)) {
```
dicej: You can simplify this to `context.value_from_bool(bool::from(value1.ct_eq(&value2)))`

dicej (approved):

crates/spin-js-engine/src/lib.rs
karthik2804: Thanks!

radu-matei (approved):
Yay!
Thanks!

PR #122, opened 2023-02-14
add support for spin KV

karthik2804:
Ci fails because it uses older version of spin.


dicej (approved):

types/lib/modules/spinSdk.d.ts
```diff
@@ -29,6 +36,9 @@ interface SpinSDK {
         smembers: (address: string, key: string) => Array<string>;
         srem: (address: string, key: string, values: Array<string>) => bigint;
     };
+    kv: {
+        open: (name: string) => KvObject;
```
radu-matei: Do we want to add a function to open the default store as well?

types/lib/modules/spinSdk.d.ts
```diff
@@ -17,6 +17,13 @@ interface HttpResponse {
     body?: ArrayBuffer;
 }
 declare type HandleRequest = (request: HttpRequest) => Promise<HttpResponse>;
+interface KvObject {
+    delete: (key: string) => void;
+    exists: (key: string) => boolean;
+    get: (key: string) => ArrayBuffer;
+    getKeys: () => Array<string>;
+    set: (key: string, value: ArrayBuffer) => void;
```
radu-matei: Do we want to make the signature here a bit more flexible?
I'm thinking about making it easier to work with strings in particular, given that (I assume) a lot of scenarios will want to work with JSON data.

karthik2804: We could add support for strings in addition to ArrayBuffers. I will add it. Just to make sure, are we talking about supporting strings in both the get and set methods?

types/lib/modules/spinSdk.d.ts
karthik2804: I am not sure it adds much value when we could open it with `spinSdk.kv.open("default")`

types/lib/modules/spinSdk.d.ts
radu-matei: Just in `set`.
Since the lower level API always returns a byte array, there is no way of knowing when to convert something.

Alternatively, we could consider `getString` and `setString`, but I'm not a huge fan of that..

karthik2804: Sounds good to me. I will add that to the `set` method.

karthik2804: This has been added.

types/lib/modules/spinSdk.d.ts
itowlson: It avoids the magic string - the compiler will tell me when (not if: when) I mistype it `defualt`.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -29,6 +29,14 @@ interface HttpResponse {
 
 type HandleRequest = (request: HttpRequest) => Promise<HttpResponse>
 
+interface KvObject {
```
itowlson: This is specifically a KV _store_ - 'object' is unnecessarily open-ended for my taste.

itowlson (changes requested):

types/lib/modules/spinSdk.d.ts
karthik2804: That makes sense. I will add it.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
karthik2804: changed it to `KvStore`.

types/lib/modules/spinSdk.d.ts
karthik2804: Added it now.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
```diff
@@ -29,6 +29,14 @@ interface HttpResponse {
 
 type HandleRequest = (request: HttpRequest) => Promise<HttpResponse>
 
+interface KvStore {
```
itowlson, line 32: `Kv` but `SDK` - do we have a principle on how we capitalise acronyms?

itowlson (commented):
This looks fine functionally but it would be good to define our standard for capitalising acronyms because it looks a bit inconsistent at the moment.

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
karthik2804: I am happy to adopt any principle we choose on this and make the required changes.

itowlson: ah good catch we already mix Http and SDK, and it would be a breaking change to change either.  Let's live with it.

itowlson (approved):

crates/spin-js-engine/src/js_sdk/modules/spinSdk.ts
karthik2804: The user actually interacts only with `spinSdk` not `SpinSDK` therefore I am not sure it would be a breaking change overall but If we do decide on a standard for this, happy to open a new PR with the required changes.

karthik2804:
closing in favour for #125

PR #123, opened 2023-02-15
Add support for redis `execute`

karthik2804:
This PR adds support for the redis `execute` method to allow executing of arbitrary redis commands.

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -14,9 +14,14 @@ use {
     spin_sdk::{
         config,
         http::{Request, Response},
+<<<<<<< HEAD
         http_component,
         key_value::Store,
         outbound_http, redis,
+=======
+        http_component, outbound_http, redis,
+        redis::RedisResult,
+>>>>>>> 0081caf (initial commit)
     },
     std::{
         collections::HashMap,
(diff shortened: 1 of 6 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -329,6 +330,74 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
     context.value_from_f64(thread_rng().gen_range(0.0_f64..1.0))
 }
 
+fn redis_exec(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
+    enum Parameter {
+        Int64(i64),
+        Binary(ByteBuf),
+    }
+
+    println!("execcuting exec");
```
dicej: Can we remove this?

dicej (approved):

crates/spin-js-engine/src/lib.rs
```diff
@@ -329,6 +330,74 @@ fn math_rand(context: &Context, _this: &Value, _args: &[Value]) -> Result<Value>
...
+                        let mut serializer = Serializer::from_context(context)?;
+                        val.serialize(&mut serializer)?;
+                        arr.append_property(serializer.value)?;
+                    }
+                }
+            }
+            return Ok(arr);
+        }
+        _ => bail!(
+            "expected a two arguments (address, key), got {} arguments",
```
dicej: This message doesn't match the actual signature.

karthik2804: Thanks for catching that.

crates/spin-js-engine/src/lib.rs
karthik2804: done.

PR #124, opened 2023-02-15
use canary spin to run tests

karthik2804:


radu-matei:
There is a database file committed to this PR.


Also, you can update the URL in the configuration action to canary.
That will continue to work and make things easier if we ever want to have cross platform actions.

karthik2804:
Thanks for pointing that out! The changes have been made.

radu-matei (approved):

PR #130, opened 2023-02-19
Update class declaration for `ResponseBuilder`

radu-matei:
This commit updates the class declaration for `ResponseBuilder` to return the explicit `ResponseBuilder` type on builder methods. This is because returning `this` does not help infer the actual type in the LSP clients used by code editors, which means chaining for the response builder was not getting auto-complete.


karthik2804 (approved):
LGTM!

PR #136, opened 2023-02-24
return null on nosuchkey

karthik2804:
Return `null` instead of an error when the key does not exist while using the `get` method on a KV store.

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -14,7 +14,7 @@ use {
     spin_sdk::{
         config,
         http::{Request, Response},
-        http_component,
+        http_component, key_value,
         key_value::Store,
         outbound_http, redis,
         redis::RedisResult,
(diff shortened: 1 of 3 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -635,10 +637,16 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
-                            Ok(serializer.value)
+                            let result = store.get(key);
+                            match result {
+                                Ok(val) => {
+                                    let mut serializer = Serializer::from_context(context)?;
+                                    val.serialize(&mut serializer)?;
+                                    Ok(serializer.value)
+                                }
+                                Err(key_value::Error::NoSuchKey) => context.null_value(),
+                                Err(e) => bail!(e),
```
dicej: `bail` might do some kind of unexpected conversion here -- could we change it to:
```suggestion
                                Err(e) => return Err(e.into()),
```

dicej (approved):

crates/spin-js-engine/src/lib.rs
karthik2804: Ah! Will change it.

PR #138, opened 2023-02-24
wrap `Store` in `Rc` prior to cloning

dicej:
Per https://github.com/fermyon/spin/pull/1200, `Store` should not have implemented `Clone` in the first place, so we wrap it in an `Rc` to avoid cloning it.

This also addresses some overdue Clippy fixes and adds Clippy and Rustfmt checks to CI.


karthik2804 (approved):
LGTM

PR #142, opened 2023-03-03
add type for KvStore to global space

### LINKED ISSUE #141: Type definitions for `KvStore`
I want to be able to write the following function:

```js
function get(kv: KvStore, id: string)...
```

However, that does not seem to be currently possible: `cannot find name KvStore`

karthik2804:
fixes #141 

This does add `KvStore` to the global namespace. One other solution would be to add it as a property under `spinSdk` and use the following to declare the type.

```js
typeof spinSdk["KvStore"]
```


radu-matei (approved):
Thanks!

LGTM

dicej (commented):
Personally, I'd prefer to stop adding non-standard things to the global namespace if possible and keep them scoped to the appropriate SDK module.

PR #156, opened 2023-04-01
add support for redis and postgres

karthik2804:
This PR adds support to both mysql and postgres. 

The examples currently use the plugin and types from `HEAD` until a new release is made.

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -16,8 +16,8 @@ use {
         http::{Request, Response},
         http_component, key_value,
         key_value::Store,
-        outbound_http, redis,
-        redis::RedisResult,
+        mysql, outbound_http, pg,
+        redis::{self, RedisResult},
     },
     std::{
         collections::HashMap,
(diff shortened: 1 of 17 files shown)


crates/spin-js-engine/src/lib.rs
```diff
@@ -16,7 +16,10 @@ use {
         http::{Request, Response},
         http_component, key_value,
         key_value::Store,
-        outbound_http, redis,
+        mysql::{self},
```
dicej:
```suggestion
        mysql,
```

crates/spin-js-engine/src/lib.rs
```diff
@@ -16,7 +16,10 @@ use {
...
+        outbound_http,
+        pg::{self},
```
dicej:
```suggestion
        pg,
```

crates/spin-js-engine/src/lib.rs
```diff
@@ -16,7 +16,10 @@ use {
...
+        redis,
```
dicej: Could combine this line with the next one, e.g. `redis::{self, RedisResult}`

crates/spin-js-engine/src/lib.rs
```diff
@@ -696,6 +699,287 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
+
+fn rdbms_param_deserializer(params: &Value, arg: &mut Vec<RdmsParameter>) -> Result<()> {
+    let mut props = params.properties()?;
+    while let Some(ref _x) = props.next_key()? {
+        let val = props.next_value()?;
+        if val.is_bool() {
+            let deserializer = &mut Deserializer::from(val.clone());
+            let temp = bool::deserialize(deserializer)?;
+            arg.push(RdmsParameter::Boolean(temp));
+        }
```
dicej:
```suggestion
        } else
```

crates/spin-js-engine/src/lib.rs
```diff
@@ -696,6 +699,287 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
+            let statement = deserialize_helper(statement)?;
+            let mut arg = vec![];
+            if params.is_array() {
+                rdbms_param_deserializer(params, &mut arg)?;
+
+                mysql::execute(
+                    &address,
+                    &statement,
+                    &arg.iter()
+                        .map(|arg| match arg {
```
dicej: Can we move this `match` statement into its own function so it can be reused by `mysql_query`?

crates/spin-js-engine/src/lib.rs
```diff
@@ -696,6 +699,287 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
+                            mysql::DbValue::Uint32(v) => context.value_from_u32(*v)?,
+                            mysql::DbValue::Uint64(v) => context.value_from_u64(*v)?,
+                            mysql::DbValue::Floating32(v) => {
+                                context.value_from_f64(v.to_owned().into())?
+                            }
+                            mysql::DbValue::Floating64(v) => context.value_from_f64(*v)?,
+                            mysql::DbValue::Str(v) => context.value_from_str(v)?,
+                            mysql::DbValue::Binary(v) => context.array_buffer_value(v)?,
+                            mysql::DbValue::DbNull => context.null_value()?,
+                            mysql::DbValue::Unsupported => context.undefined_value()?,
```
dicej: I'm thinking we should return an error here instead of using `undefined`.

crates/spin-js-engine/src/lib.rs
```diff
@@ -696,6 +699,287 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
+            let statement = deserialize_helper(statement)?;
+            let mut arg = vec![];
+            if params.is_array() {
+                rdbms_param_deserializer(params, &mut arg)?;
+
+                pg::execute(
+                    &address,
+                    &statement,
+                    &arg.iter()
+                        .map(|arg| match arg {
```
dicej: As with `mysql_execute`, let's deduplicate the `match` statements here and in `postgres_query`.

crates/spin-js-engine/src/lib.rs
```diff
@@ -696,6 +699,287 @@ fn open_kv(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
...
+                            pg::DbValue::Uint32(v) => context.value_from_u32(*v)?,
+                            pg::DbValue::Uint64(v) => context.value_from_u64(*v)?,
+                            pg::DbValue::Floating32(v) => {
+                                context.value_from_f64(v.to_owned().into())?
+                            }
+                            pg::DbValue::Floating64(v) => context.value_from_f64(*v)?,
+                            pg::DbValue::Str(v) => context.value_from_str(v)?,
+                            pg::DbValue::Binary(v) => context.array_buffer_value(v)?,
+                            pg::DbValue::DbNull => context.null_value()?,
+                            pg::DbValue::Unsupported => context.undefined_value()?,
```
dicej: As above, I'd like to return an error instead of using `undefined` here.

dicej (changes requested):

karthik2804:
@dicej implemented the changes in the latest push.

dicej (approved):

Issue #158, opened 2023-04-02
Top level `.d.ts` file containing all SDK types


radu-matei:
Currently, the SDK contains a mix of built-in types (such as `utils.Router`, `spinSdk`, or `ResponseBuilder`) and types that can be inspected in the TypeScript source.

This creates confusion when some types need to be explicitly imported, while other don't.

This issue asks for a type definition file that contains all the types ever used in the SDK, so all types would be imported from the type definition file.

cc @karthik2804

karthik2804:
Looking into it. Just a little bit confused on what you mean by types that can be inspected from TypeScript source.

radu-matei:
Sorry, I meant the types that can be found in the `sdk.ts` source file in this repo.

karthik2804:
All the types are defined here https://github.com/fermyon/spin-js-sdk/tree/main/types/lib in this same repo (includes spinSdk and others). I will look to make it such that the types have to be explicitly imported.

Along those lines, there is a case of pure JS users where the `spinSdk` module will still be global level without any explicit imports.

One solution is to make it such that we move away from just providing a types file and make it an actual library with an implementation that wraps the internal types into implementations of its own that can be explicitly imported (this would require a little bit of restructuring of how things are currently done). I can look to make a simple PoC

karthik2804:
As a follow-up to the previous, I believe that making just the types explicitly imported can be done by manually straightforwardly writing the types file. 

meaning both of the following will be valid and compile and execute 
```js
import { spinSdk} from "spin-sdk"
..... {
let a = spinSdk
}
```
```js
... {
//@ts-ignore
let a = spinSdk
}
```

karthik2804:
We now have a new SDK with much better type support.

PR #159, opened 2023-04-02
WIP: POC for new sdk js lib implementation

karthik2804:
This is a very work-in-progress implementation of the SDK lib. This PR attempts to replace the current `spin-sdk` package which hosts only type to a package that provides implementations as well. This allows for there to be an explicit import of types in both ts/js as per #158.

Currently, there are examples for both ts/js In 
- `examples/javascript/sdk-poc-js`
-  `examples/javascript/sdk-poc-ts`

The advantage to this is that helper functions can be added without a need for there to be a new build of the plugin along with allowing us to create explicit imports allowing for the removal of magic imports.

The one downside to this approach is that there will be slightly more work maintaining both the code that is internal to the SDK as well as this package.

There definitely needs to be a lot more of the cleanup if this is an approach we want to pursue.

Thoughts?


spin-sdk/src/modules/spinSdk.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+
+/**
+ * Sdk module that provides access to spin features
+ */
+const spinSdk: SpinSdk = __internal__.spin_sdk
+spinSdk.utils = utils
+spinSdk.Router =  () => {
```
radu-matei, line 99: nit: could we have `Router` as a top-level import from the SDK, instead of it being namespaced as `spinSdk.Router`?

radu-matei (approved):
This is looking great, LGTM!
(post this merging we need to update the docs for the JS SDK as well)

dicej (approved):

PR #165, opened 2023-05-10
Change template .wasm file to use project-name parameter

mikkelhegn:
It makes sense to call the .wasm output file using the template name, as this is the convention we have for other templates.

A potential addition could be to parameterize this to use `{{project-name | kebab_case}}.wasm`, like the component id.

Marking this a draft until we agree on whether this should be parameterized.

### DIFF
--- a/templates/http-js/content/package.json
+++ b/templates/http-js/content/package.json
@@ -4,7 +4,7 @@
   "description": "{{project-description}}",
   "main": "index.js",
   "scripts": {
-    "build": "npx webpack --mode=production && mkdir -p target && spin js2wasm -o target/spin-http-js.wasm dist/spin.js",
+    "build": "npx webpack --mode=production && mkdir -p target && spin js2wasm -o target/{{project-name | kebab_case}}.wasm dist/spin.js",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "keywords": [],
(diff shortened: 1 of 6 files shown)


karthik2804:
I think parameterizing the name of the output file sounds like a good idea to me. Allowing for easier identification of the purpose of the wasm module. The `package.json` needs to be updated for that.

itowlson:
Some templates (e.g. Rust) parameterise the name of the output file.  Others (e.g. Go) do not.  I like to imagine it depends on what is convenient in the language/toolchain, though it is probably really historical accident inherited from whoever wrote the example the template was adapted from.  In any case, parameterising the Wasm file name is definitely _not_ an established convention, and template authors should _not_ feel any obligation or pressure to do it.

itowlson:
And... to be clear... that is not a "don't parameterise"... that is a "parameterise if it is more convenient or if you prefer it that way."

mikkelhegn:
I'd love for this to be a convention, it just makes `spin.toml` much easier to read, if the wasm modules don't have generic names. but get your point @itowlson :-)

mikkelhegn:
BTW - the handle bar templating works really nicely 👍🏻

karthik2804 (approved):
LGTM!

PR #166, opened 2023-05-10
add plugin manifest generation script

karthik2804:
Adds CI to generate canary plugins and now generates the plugin release manifests as well.


dicej (approved):

.github/workflows/release.yaml
```diff
@@ -9,9 +11,32 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  reset_canary_release:
```
dicej, line 14: @karthik2804 We should probably _not_ run this if we're building a release from a tag, right?  It looks like it's set up to run for both tag and `main` pushes.

.github/workflows/release.yaml
```diff
@@ -159,12 +188,24 @@ jobs:
           file: _dist/js2wasm-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}.tar.gz
           tag: ${{ github.ref }}
 
-  checksums:
+      - name: upload binary to Github release (canary)
```
dicej, line 191: Likewise, this should be gated on `if: github.ref == 'refs/heads/main'`

.github/workflows/release.yaml
karthik2804: Yea. I need to fix it. Did not have that enabled because of testing on a different branch. Will make a follow up PR to fix it.

PR #167, opened 2023-05-12
dont produce canary artifacts on release

karthik2804:


karthik2804:
We will not need this anymore once we move to the `componentize-js` based SDK

PR #170, opened 2023-06-21
Improve build docs and add vs code settings

rylev:
This would have saved me some headache when getting set up in this project.


dicej (approved):

rylev:
Looks like something really funny is happening inside of the `quickjs_wasm_rs::Context` initialization. Seems like we're successfully creating a JS runtime but then creating the inner Context fails with a "out of bounds memory access". I can't reproduce this locally.

radu-matei (approved):

karthik2804:
@rylev I have been running into this as well. The solution seems to be to upgrade `wizer`

PR #171, opened 2023-06-22
Add sqlite support

rylev:
This adds support for sqlite to the JS SDK along with a very simple example app using the SDK.


karthik2804:
We would need to add some "typings" for the TS compiler but that would require a refactor once the [SDK refactor](https://github.com/fermyon/spin-js-sdk/pull/159) gets merged.

I can add the type of support as a follow-up PR.

karthik2804 (approved):

karthik2804 (approved):
LGTM

PR #176, opened 2023-07-10
compile releases with statically with openssl

karthik2804:


.github/workflows/release.yaml
```diff
@@ -42,20 +42,24 @@ jobs:
...
           - {
-              os: "ubuntu-latest",
+              os: "ubuntu-20.04",
               arch: "amd64",
               wasiSDK: "linux",
               extension: "",
-              buildArgs: "",
+              # Ubuntu 22.04 no longer ships libssl1.1, so we statically
+              # link it here to preserve release binary compatibility.
+              buildArgs: "\"--features openssl/vendored\"",
```
lann, line 51: Is `spin-js-sdk` actually using openssl? :thinking:

lann: It doesn't appear in the `Cargo.lock`...

karthik2804: I just looked as well 😅

karthik2804: I am going to close this PR.

PR #179, opened 2023-07-12
add version back to cli options

karthik2804:
The switch from `structopt` to `clap` seems to have introduced a regression in the CLI arguments for version.


rylev (approved):

rajatjindal:
can we do a new release with this fix?

PR #184, opened 2023-08-18
add spin-kv-ts example

karthik2804:


examples/typescript/spin_kv/README.md
```diff
@@ -0,0 +1,38 @@
+## Spin KV example
```
itowlson:
```suggestion
## Key-value storage example
```

examples/typescript/spin_kv/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This is a simple example showcasing using Spin key-value with spin-js-sdk using typescript.
```
itowlson:
```suggestion
This is a simple example showcasing using Spin key-value storage with TypeScript.
```

examples/typescript/spin_kv/README.md
```diff
@@ -0,0 +1,38 @@
...
+###  Building and Running
+
+To build the app run the following commands:
+
+```bash
+$ npm install
+$ spin build --up
+```
+
+The application can now receive requests on http://localhost:3000:
```
itowlson, line 22: What sort of requests - looks like key is in the URL and POST with body to set, GET to read, DELETE to delete?  PUT?

itowlson (changes requested):
Just some slight burnishing of grammar if this is to be an official sample!

examples/typescript/spin_kv/README.md
karthik2804: There is no put request, the mapping between the request method and KV operation is as follows: 

GET -> kv.get(request.url)
POST -> kv.set(request.uri, body)
HEAD -> kv.exists(request.uri)
DELETE -> kv.delete(request.uri)

karthik2804: Follows the same structure as https://github.com/fermyon/spin/blob/main/examples/rust-key-value/src/lib.rs

itowlson: It would be great to have that info in the README though (where readers of the sample will see it) rather than in a GH comment... _grin_

itowlson (approved):
one smol request but no blockers

PR #185, opened 2023-09-01
update sqlite return types to be more ergonomic

### LINKED ISSUE #183: SqliteReturn type uses tuple with union type instead of an array type
The SqliteReturn type uses a tuple for the rows:
```ts
interface SqliteReturn {
    columns: string[],
    rows: [
        [SqliteValue]
    ]
}
```

This makes parsing results more complicated when you need to manipulate it as an array (when you expect more than one row.
I suggest changing the response type to be:

```ts
interface SqliteReturn {
    columns: string[],
    rows: SqliteValue[]
}
```
Or maybe something more suitable for different types, following the example of Planetscale:
https://github.com/planetscale/database-js/blob/488e48b04cfb25b001050d6dbeebe8dd9b0a6f65/src/index.ts#L7 
```ts
type Row = Record<string, any> | any[]
```

karthik2804:
Changes the type to a hashmap of the column names. 

fixes #183

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -745,24 +745,24 @@ fn open_sqlite(context: &Context, _this: &Value, args: &[Value]) -> Result<Value
 
                 let mut serializer = Serializer::from_context(context)?;
                 let columns = result.columns;
-                columns.serialize(&mut serializer)?;
-                let js_columns = serializer.value;
-
                 let js_rows = context.array_value()?;
                 for row in result.rows {
-                    let js_row = context.array_value()?;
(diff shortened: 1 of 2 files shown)


rylev (changes requested):
This is technically a breaking change. This feature is in private preview so perhaps that's fine, but I'd like to make sure we socialize this change widely before accepting it. If everyone is fine with the breaking change, this looks good to me.

dicej (approved):
Given that both the SDK itself and the SQLite feature are still considered "experimental", I think it's reasonable to make a breaking change to improve the API.

rylev (approved):

PR #187, opened 2023-09-04
Add llm

karthik2804:
This add support in the JS SDK for the `Llm` interface in reference to the Spin PR fermyon/spin#1731

### DIFF
--- a/crates/spin-js-engine/src/lib.rs
+++ b/crates/spin-js-engine/src/lib.rs
@@ -16,7 +16,7 @@ use {
         http::{Request, Response},
         http_component, key_value,
         key_value::Store,
-        mysql, outbound_http, pg,
+        llm, mysql, outbound_http, pg,
         redis::{self, RedisResult},
         sqlite,
     },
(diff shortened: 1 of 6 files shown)


radu-matei (approved):
Tests are failing here.
Otherwise, the PR looks good to me; feel free to merge once CI is fixed.

crates/spin-js-engine/src/lib.rs
```diff
@@ -1045,6 +1044,132 @@ fn postgres_query(context: &Context, _this: &Value, args: &[Value]) -> Result<Va
...
+struct InferencingOption {
+    max_tokens: u32,
+    repeat_penalty: f32,
+    repeat_penalty_last_n_token_count: u32,
+    temperature: f32,
+    top_k: u32,
+    top_p: f32,
+}
+
+fn llm_inference_with_options(context: &Context, _this: &Value, args: &[Value]) -> Result<Value> {
```
rylev, line 1095: Are we sure we want this? We do this in Rust because Rust doesn't have optional args, but in JS it feels much more idiomatic to just have one function.

rylev (approved):

crates/spin-js-engine/src/lib.rs
rylev: Seeing now that the JS code differentiates these two functions. We should clean this up in a follow-up PR.

PR #197, opened 2023-09-25
Build on Ubuntu 20.04

### LINKED ISSUE #182: Build Linux binary on Ubuntu 20.04
Currently, we're building on `ubuntu-latest`, which prevents the binary from running on older versions.  We should match Spin's release.yml and use `ubuntu-20.04` instead.

See https://github.com/fermyon/spin/issues/1676 for more details.

itowlson:
Fixes #182.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         config:
           - {
-              os: "ubuntu-latest",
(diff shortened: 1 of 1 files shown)


lann (approved):

dicej (approved):
Thanks, Ivan!

karthik2804 (approved):

PR #200, opened 2023-10-17
Add key-value sample for JavaScript

### LINKED ISSUE #164: Add example using the kv api
Having an example using the kv api would be a great addition.

thorstenhans:
This PR adds an example to illustrate how developers can use key-value APIs using the `spin-js-sdk`

closes #164

### DIFF
--- a/examples/javascript/kv/src/index.js
+++ b/examples/javascript/kv/src/index.js
@@ -1,0 +1,132 @@
+import { Router, Kv } from '@fermyon/spin-sdk'
+
+const en = new TextEncoder()
+const de = new TextDecoder()
+
+let router = new Router()
+
+router.get("/", getAll)
+router.get("/json/:key", ({ params }) => getJson(params.key))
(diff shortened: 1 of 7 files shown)


examples/javascript/kv/src/index.js
```diff
@@ -0,0 +1,132 @@
+import { Router, Kv } from '@fermyon/spin-sdk'
+
+const en = new TextEncoder()
+const de = new TextDecoder()
```
karthik2804: Just a tiny nit - could this be renamed as `encoder` and `decoder` to keep it consistent with the other examples?

karthik2804 (approved):
This is looking great. Thanks!

I just have one small nit.

examples/javascript/kv/src/index.js
thorstenhans: Sure will fix it

karthik2804:
Thank you once again for tackling this @ThorstenHans!

PR #202, opened 2023-10-18
Add samples for built-in HTTP router (JS & TS)

### LINKED ISSUE #163: Add examples of using the built-in router
An example using the built-in router would be a great addition.

thorstenhans:
This PR adds two simple example applications to demonstrate the usage of the built-in router provided by `spin-js-sdk`.

fixes #163


vdice (approved):
Thanks for adding these @ThorstenHans!

PR #213, opened 2023-11-13
add some initial docs

karthik2804:


spin-sdk/src/modules/spinSdk.ts
```diff
@@ -2,16 +2,62 @@ import { router, routerType } from "./router"
...
+     * @returns - A value string for the given key
+     */
     get(arg0: string): string
 }
 
 interface KvStore {
+    /**
+     * Delete the key
+     * 
+     * @param arg0 - The name of the key
```
itowlson:
```suggestion
     * @param key - The name of the key
```

(looks like this might be throughout)

spin-sdk/src/modules/spinSdk.ts
```diff
@@ -177,11 +223,16 @@ declare global {
 }
 
 const kv = {
+    /**
+     *
```
itowlson, line 227: Is this meant to have a method description?

spin-sdk/src/modules/spinSdk.ts
```diff
@@ -177,11 +223,16 @@ declare global {
...
 
 const kv = {
+    /**
+     * 
+     * @param name - The name of the KV store to open
+     * @returns A KV store handle
+    */
     open: (name: string) => {
         let store = __internal__.spin_sdk.kv.open(name)
         store.getJson = (key: string) => {
```
itowlson, line 233: We can't add docstrings to these members?

itowlson (approved):
Some tweaks but not blockers, and better to have this done!  Thanks for this and for getting it done so super quickly!

spin-sdk/src/modules/spinSdk.ts
karthik2804: Not sure I follow.

itowlson: The object returned by `open()` has functions on it.  Is there a way for someone using the docs to know what those functions are, or what their parameters are?

(Or is the return value of `open()` meant to be an opaque handle and these functions not meant to be called by application code?)

karthik2804: One is able to click on the return type to know the functions that can be called. The return type here is `KvStore`. The `setJson` and `getJson` are added to satisfy that interface.

Issue #209, opened 2023-11-05
Scripts don't support Windows, blocking Getting Starting guide


markenzie:
Hi, I am following the getting started guide and I may be doing something wrong.

From my issues, it seems due to this line

https://github.com/fermyon/spin-js-sdk/blob/c76bec67a15a2dfaf206bf2da3ed99a5c487c714/examples/javascript/kv/package.json#L7

On Windows this creates a -p folder as that is not a valid option.

```
C:\Users\me\Desktop\projects\test>ls
-p  README.md  dist  node_modules  package-lock.json  package.json  spin.toml  src  target  webpack.config.js
```

This puts me into a failed state when trying to build or deploy for a second time:

```
C:\projects\enzie-test>spin build
Building component test with `npm run build`

> enzie-test@1.0.0 build
> npx webpack --mode=production && mkdir -p target && spin js2wasm -o target/enzie-test.wasm dist/spin.js

asset spin.js 1.87 KiB [compared for emit] (name: main)
runtime modules 670 bytes 3 modules
./src/index.js 183 bytes [built] [code generated]
webpack 5.89.0 compiled successfully in 78 ms
A subdirectory or file -p already exists.
Error occurred while processing: -p.
A subdirectory or file target already exists.
Error occurred while processing: target.
Error: Build command for component enzie-test failed with status Exited(1)

C:\projects\enzie-test>spin deploy
Error: The system cannot find the file specified. (os error 2)
```

Just so you guys are aware. I haven't entirely figured out the correct process to make this work for me either.  It appears even commands like `spin login` give the same error as spin deploy above, both inside and outside of any working directory. 

Am I doing something really wrong here?

karthik2804:
Hello! Thanks for reporting this. The build error is due to `mkdir` on windows now supporting the `-p` argument. There are a couple of solutions to it. 
1. The first would be to set up npm to run the scripts using bash (https://developer.fermyon.com/spin/v2/javascript-components#a-quick-note-about-the-npm-script-and-windows).
```bash
npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"
```
2. Modify the script to manually create the `target` directory and edit the build script to be. 
```json
"build": "npx webpack --mode=production && spin js2wasm -o target/kv.wasm dist/spin.js", 
```

[...]

PR #214, opened 2023-11-28
Windows mkdir fix 

oferh:
fix for #209 
replace `mkdir -p` with the mkdirp package

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1667,7 +1667,7 @@ dependencies = [
 
 [[package]]
 name = "spin-js-cli"
-version = "0.6.0"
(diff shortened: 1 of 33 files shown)


dicej (approved):

dicej:
Looks great -- thanks for this, @oferh!  Would you mind signing your commit?  We have a branch rule set up that requires commits to be signed prior to merging.

karthik2804:
This is great! Thanks. I think the examples will also need the same change. I can update the examples as a follow-up.

oferh:
@dicej signed the commit.
@karthik2804 - added fix for examples and test

PR #222, opened 2024-04-25
fix import name to match generated wit

karthik2804:
The generated function is `generateEmbedding` based on the `wit`


jpflueger (approved):
LGTM, thanks for this!

Issue #204, opened 2023-10-26
Examples: Update planetscale to use variables


vdice:
The [planetscale js](https://github.com/fermyon/spin-js-sdk/tree/main/examples/javascript/planetscale) and [planetscale ts](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/planetscale) examples are good candidates to refactor to use [Application Variables](https://developer.fermyon.com/spin/variables) instead of the `<...>` placeholders in the index.js/ts files.

PR #229, opened 2024-05-05
Added variables to planetscale example

VamshiReddy02:
Ref: #204 
Let me know if there are any changes to be made.


examples/typescript/planetscale/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
 
 import { connect } from '@planetscale/database'
 
 const config = {
-   host: '<host>',
-   username: '<username>',
-   password: '<password>'
+   host: Config.get("host"),
+   username: Config.get("username"),
+   password: Config.get("password")
```
karthik2804, line 8: These will need to be moved inside the hander. It will error out when used in the top level.

karthik2804 (changes requested):

karthik2804:
Closing as Old SDK is no longer under active development.

PR #230, opened 2024-05-06
mark all modules as sideeffect free

karthik2804:
Helps with treeshaking, none of the exports have any side effects so marking them as not having side effects.


karthik2804:
@radu-matei can we merge this? 

ref: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

radu-matei (approved):

PR #234, opened 2024-05-14
bundle default set of wit files with sdk

karthik2804:
This allows to not have to specify the wit files in every application as long as they target the `http` or `redis` triggers.I have not intentionally updated the templates as the templates install the version of the package from npm.


bin/j2w.mjs
```diff
@@ -35,6 +37,13 @@ const args = yargs(hideBin(process.argv))
 
 const src = args.input;
 const source = await readFile(src, 'utf8');
+
+// Check if a non default wit directory is supplied
+if (!args.witPath) {
+    console.log("using default wit")
```
karthik2804: Do we think this is something that is useful information for the user?

radu-matei: I think it's useful to log the other branch.

karthik2804: updated.

radu-matei (approved):

PR #235, opened 2024-05-14
add webpack plugin to manage the externals configuration

karthik2804:
Once this is merged, the webpack config on apps can become something like the following where the plugin takes care of injecting the externals. 

```js
const path = require('path');
const SpinSDKPlugin = require("@fermyon/spin-sdk/plugin/webpack")

module.exports = {
    ...
    plugins: [
        new SpinSDKPlugin()
    ],
    ...
};
```

### DIFF
--- a/plugins/webpack/index.js
+++ b/plugins/webpack/index.js
@@ -1,0 +1,25 @@
+class SpinSdkPlugin {
+    constructor() {
+        this.externals = {
+            "wasi:http/types@0.2.0": "wasi:http/types@0.2.0",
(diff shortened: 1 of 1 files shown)


plugins/webpack/index.js
```diff
@@ -0,0 +1,25 @@
+class SpinSDKPlugin {
```
radu-matei: The camel case naming conventions we are using elsewhere means we should name this `SpinSdkPlugin`

karthik2804: Updated.

radu-matei (approved):

PR #238, opened 2024-05-15
remove .wit folder from the template and update build command

karthik2804:


karthik2804:
~need to fix plugin imiport in webpack.~

done

radu-matei (approved):

PR #239, opened 2024-05-15
Add ci sdk v2

karthik2804:
This PR adds `prettier` for code formatting as well as creates a test app that currently does nothing more than just build and return simple responses. It lays the foundation to add tests as HTTP routers once `fetch` is enabled in a similar fashion to the older SDK.

PS: most of the changes are from prettier for consistent formatting.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,9 +5,10 @@ on:
     branches: [ feat/sdk-v2 ]
   pull_request:
     branches: [ feat/sdk-v2 ]
+  workflow_dispatch:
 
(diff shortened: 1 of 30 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,41 @@
+name: Test
+
+on:
+  push:
+    branches: [ feat/sdk-v2 ]
```
radu-matei, line 5: TODO: remember to update this once we switch this to be the main branch.

radu-matei (approved):

PR #236, opened 2024-05-14
[sdk-v2] add Vue example

bacongobbler:
This adds a Vue.js example to the 2.0.0 release of the javascript SDK.

PR #233, opened 2024-05-13
[sdk-v2] add NuxtJS example

bacongobbler:

PR #244, opened 2024-06-03
update SDK to use fetchEvent

karthik2804:
The templates have not been updated since there is a release of the SDK and we cannot use a git reference as this repository does not check in artifacts. 

Still a work in progress because it needs a little more testing as well as needs to be updated once a fix for https://github.com/bytecodealliance/StarlingMonkey/pull/52#issuecomment-2145301998 is pushed.


bacongobbler:
Note that we will need to update the Vue and Nuxt examples in #236 and #233.

bacongobbler:
Attempting to compile an app with this new version of the SDK failed with the following:

```
Exception while pre-initializing: (new TypeError("util$1.types is undefined", "spin.js", 166026))

Error: the `componentize.wizer` function trapped

Caused by:
    0: error while executing at wasm backtrace:
           0: 0x73dcbf - <unknown>!<wasm function 12297>
           1: 0x73d15c - <unknown>!<wasm function 12071>
           2: 0x4bd7ae - <unknown>!<wasm function 1733>
           3: 0x2fb96a - <unknown>!<wasm function 553>
    1: Exited with i32 exit status 1
file:///home/bacongobbler/code/github.com/fermyon/spinoculars-api/node_modules/@bytecodealliance/componentize-js/src/componentize.js:217
    throw new Error(err);
          ^

Error: Failed to initialize the compiled Wasm binary with Wizer:
Wizering failed to complete
    at componentize (file:///home/bacongobbler/code/github.com/fermyon/spinoculars-api/node_modules/@bytecodealliance/componentize-js/src/componentize.js:217:11)
    at async file:///home/bacongobbler/code/github.com/fermyon/spinoculars-api/node_modules/@fermyon/spin-sdk/bin/j2w.mjs:48:23

Node.js v20.13.1
Error: Build command for component app failed with status Exited(1)
```

README.md
```diff
@@ -0,0 +1,37 @@
+# Spin JS SDK
+
+This is an SDK for Javascript and Typescript based on [componentize-js](https://github.com/bytecodealliance/ComponentizeJS).
```
tschneidereit: While the package name is `componentize-js`, the project name is `ComponentizeJS`, so we should reflect that.

```suggestion
This is an SDK for Javascript and Typescript based on [ComponentizeJS](https://github.com/bytecodealliance/ComponentizeJS).
```

README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Running the app
+
+```bash
+spin up
+```
+
+## Note: Installing the package
+
+Currently pre-release versionf are published on NPM and can be installed using the following command:
```
tschneidereit:
```suggestion
Currently pre-release versions are published on NPM and can be installed using the following command: 
```

README.md
```diff
@@ -0,0 +1,37 @@
...
+
+[Spin](https://github.com/fermyon/spin) is a prerequisite. 
+
+The templates can be installed with the following command:
+
+```bash
+spin plugin install --update --git https://github.com/fermyon/spin-js-sdk --branch feat/sdk-v2
+```
+
+## Creating and Building a new app=
```
tschneidereit: We should keep that casing consistent for headlines.

```suggestion
## Creating and building a new app
```

templates/http-ts/content/src/spin.ts
```diff
@@ -1,10 +1,28 @@
...
+
+    await handler(event.request, res)
 }
 
-export const incomingHandler = new App()
+// Keep wizer happy during pre-init. Should go away
+// oncehttps://github.com/bytecodealliance/ComponentizeJS/issues/114 is resolved
+export const incomingHandler = {
+    handle() { }
+}
```
tschneidereit:
```suggestion
}

```

templates/http-ts/content/src/spin.ts
```diff
@@ -1,10 +1,28 @@
...
+import { ResponseBuilder } from "@fermyon/spin-sdk";
 import { handler } from ".";
 
-class App extends HttpHandler {
-    handleRequest(req: HttpRequest, res: ResponseBuilder): Promise<void> {
-        return handler(req, res)
-    }
+//@ts-ignore
+addEventListener('fetch', (event: FetchEvent) => {
+    handleEvent(event)
```
tschneidereit:
```suggestion
    handleEvent(event);
```

templates/http-ts/content/src/spin.ts
```diff
@@ -1,10 +1,28 @@
...
+
+    let resolve: any, reject: any;
+    let responsePromise = new Promise(async (res, rej) => {
+        resolve = res;
+        reject = rej;
+    });
+    //@ts-ignore
+    event.respondWith(responsePromise);
+
+    let res = new ResponseBuilder(resolve)
```
tschneidereit:
```suggestion
    let res = new ResponseBuilder(resolve);
```

templates/http-ts/content/package.json
```diff
@@ -15,11 +15,9 @@
...
-        "webpack-cli": "^4.10.0",
-        "@bytecodealliance/componentize-js": "^0.8.3",
-        "yargs": "^17.7.2"
+        "webpack-cli": "^4.10.0"
     },
     "dependencies": {
-        "@fermyon/spin-sdk": "^2.0.0-alpha.1"
+        "@fermyon/spin-sdk": "^2.0.0-alpha.2"
     }
 }
```
tschneidereit:
```suggestion
}

```

templates/http-ts/content/src/index.ts
```diff
@@ -1,6 +1,6 @@
...
+import { ResponseBuilder } from "@fermyon/spin-sdk";
 
-export async function handler(req: HttpRequest, res: ResponseBuilder) {
-    console.log("Request: ", req.uri);
-    res.send("Hello, Universe!");
-}
+export async function handler(req: Request, res: ResponseBuilder) {
+    console.log(req)
+    res.send("hello universe")
+}
```
tschneidereit: In internal code, I'm a bit less concerned with consistent style, but for the template we should keep things like semicolon use consistent.

```suggestion
    console.log(req);
    res.send("hello universe");
}

```

src/inboundHttp.ts
```diff
@@ -43,59 +43,57 @@ export class ResponseBuilder {
...
           headers: this.headers,
           status: this.statusCode,
         }),
       );
-      this.hasWrittenHeaders = true;
-      return;
     }
-    this.streamController.pump(convertToUint8Array(value));
+    this.internalWriter!.write(contents);
+    this.hasWrittenHeaders = true;
```
tschneidereit: This should be set inside the `if` block above, after the call to `this.resolveFunction`, right?

src/inboundHttp.ts
```diff
@@ -43,59 +43,57 @@ export class ResponseBuilder {
...
+    return;
   }
   end() {
     if (this.hasSentResponse) {
       throw new Error('Response has already been sent');
     }
+    if (!this.internalWriter) {
+      throw new Error(
+        '`res.end()` called before `res.write()` - Consider using `res.send()` instead',
+      );
```
tschneidereit: I think we should either not make this an error at all, or change the message to be a bit clearer. I might lean towards the former, or specifically to make `end` call `send` implicitly if no streaming response is active: it's not "wrong" after all, just a little bit less efficient.

But in case you'd like to continue treating it as one, perhaps something like this?
```
ResponseBuilde: `end()` called without a response ever having been started. Use `send(body)` to send a non-streaming response, or write one or more chunks of the response's body using `write(chunk)` before calling `end()`.

tschneidereit (approved):
Looks good to me! I left a few comments which should be addressed, but once those are in, this looks ready

src/inboundHttp.ts
karthik2804: If it is in reference to the `hasWrittenHeaders`, then it is correct in either place but inside the `if` block removes the redundant setting. I will move it in.

src/inboundHttp.ts
karthik2804: I like the idea of not erroring, I will do that.

src/inboundHttp.ts
tschneidereit: it is in reference to that, yes. And agreed: it doesn't lead to wrong behavior where it is now, but I would argue that it's still not entirely correct in terms of intent. That doesn't really matter now, but could if some later refactoring changes things in ways that don't take it into account.

karthik2804: That makes sense, I have made the changes.

PR #251, opened 2024-06-25
catch exceptions in handler function and wait until handler completes

karthik2804:
This PR adds a `try...catch` to the template so that any unhandled exceptions are caught and returned as an error instead of the generic error along the lines of `event loop error - both task and job queues are empty, but expected operations did not resolveInternal error.`

This PR also modifies the template so that there is an `event.waitUntill` on the handler function so that applications can do some work even after the response is sent.

### DIFF
--- a/templates/http-ts/content/src/spin.ts
+++ b/templates/http-ts/content/src/spin.ts
@@ -19,13 +19,15 @@ async function handleEvent(event: FetchEvent) {
     let res = new ResponseBuilder(resolve);
 
     try {
-        // In case you want to do sonme work after the response is sent
+        // In case you want to do some work after the response is sent
         // uncomment the line below and comment out the line with 
-        // `await handler(event.request, res)
+        // await handler(event.request, res)
         // event.waitUntil(handler(event.request, res))
         await handler(event.request, res)
     } catch (e: any) {
-        res.status(500)
-        res.send(`error in handler: ${e}`)
+        if (res.getStatus() == 200) {
+            res.status(500);
+        }
+        res.send(`error in handler: ${e}`);
     }
 }
--- a/templates/http-ts/content/src/spin.ts
+++ b/templates/http-ts/content/src/spin.ts
@@ -18,5 +18,14 @@ async function handleEvent(event: FetchEvent) {
 
     let res = new ResponseBuilder(resolve);
 
-    await handler(event.request, res)
+    try {
+        // In case you want to do sonme work after the response is sent
+        // uncomment the line below and comment out the line with 
+        // `await handler(event.request, res)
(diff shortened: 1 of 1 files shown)


karthik2804:
@tschneidereit I have made the change as discussed.

templates/http-ts/content/src/spin.ts
```diff
@@ -18,7 +18,16 @@ async function handleEvent(event: FetchEvent) {
 
     let res = new ResponseBuilder(resolve);
 
-    await handler(event.request, res)
+    try {
+        // In case you want to do sonme work after the response is sent
+        // uncomment the line below and comment out the line with 
+        // `await handler(event.request, res)
+        // event.waitUntil(handler(event.request, res))
```
karthik2804: happy to replace this with a boolean and `if...else` statement if that would be cleaner.

templates/http-ts/content/src/spin.ts
```diff
@@ -18,7 +18,16 @@ async function handleEvent(event: FetchEvent) {
 
     let res = new ResponseBuilder(resolve);
 
-    await handler(event.request, res)
+    try {
+        // In case you want to do sonme work after the response is sent
```
radu-matei: typo

templates/http-ts/content/src/spin.ts
```diff
@@ -18,7 +18,16 @@ async function handleEvent(event: FetchEvent) {
...
+        // uncomment the line below and comment out the line with 
+        // `await handler(event.request, res)
```
radu-matei: extra '`' symbol here

templates/http-ts/content/src/spin.ts
```diff
@@ -18,7 +18,16 @@ async function handleEvent(event: FetchEvent) {
...
+        // event.waitUntil(handler(event.request, res))
+        await handler(event.request, res)
+    } catch (e: any) {
+        res.status(500)
+        res.send(`error in handler: ${e}`)
+    }
 }
 
 // Keep wizer happy during pre-init. Should go away
```
radu-matei: missing space on the line below

radu-matei: The issue linked here is closed -- is this still needed?

karthik2804: I will clean it up. Thoughts on the comment vs using a `if...else` if you want to continue processing after response has been sent back?

radu-matei: I don't really have a preference.

karthik2804: leaving it as  a comment, since it is a template, we can change it if we feel the need for it.

PR #252, opened 2024-07-09
add drizzle orm example

karthik2804:


radu-matei:
Could you show this with a real database, please?

karthik2804:
@radu-matei I will update it.

karthik2804:
@radu-matei updated the example to use Spin's SQLite database.

PR #254, opened 2024-07-18
update the template and add more examples on using spin host components

karthik2804:
This should not be merged until this [PR](https://github.com/bytecodealliance/ComponentizeJS/pull/122) lands and the SDK is updated to use a newer release of ComponentizeJS and a new release for the SDK is created.

Todo:
- [ ] update `package.json` in all the examples with the latest spin-sdk release
- [x] add `knitwit.json` to all examples/templates.


karthik2804:
@radu-matei can we merge this and then update the package.json once we have a new alpha?

PR #261, opened 2024-07-22
update docs to have error type information as well

karthik2804:


karthik2804:
I am wondering if this should just be built in CI and deployed to pages as an artifact.

radu-matei:
Oh yeah, doing this in CI would be great.

karthik2804:
I assume we still want to keep the V1 docs as well. I have to take a look at how that will work.

karthik2804:
Docs pushed to a seperate branch 
https://github.com/fermyon/spin-js-sdk/tree/gh_pages

PR #267, opened 2024-07-30
add ci to test template

karthik2804:


.github/workflows/test.yaml
```diff
@@ -38,4 +38,40 @@ jobs:
...
+
+    steps:
+        - uses: actions/checkout@v3
+
+        - name: Use Node.js 20
+          uses: actions/setup-node@v3
+          with:
+            node-version: 20
+
+
```
vdice: nit: extra newline/whitespace 😅

.github/workflows/test.yaml
```diff
@@ -38,4 +38,40 @@ jobs:
...
             cd test
-            ./test.sh
\ No newline at end of file
+            ./test.sh
+  
+  test_template:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
vdice, line 43: Out of curiousity, is there anything os-specific about the test template?  Or are we mostly testing spin/npm on each OS?

.github/workflows/test.yaml
```diff
@@ -38,4 +38,40 @@ jobs:
...
+        - uses: actions/checkout@v3
+
+        - name: Use Node.js 20
+          uses: actions/setup-node@v3
+          with:
+            node-version: 20
+
+            
+        - name: Install Spin
+          uses: rajatjindal/setup-actions/spin@main
```
vdice: Can we use the official [spin/setup](https://github.com/fermyon/actions?tab=readme-ov-file#install-spin-cli-and-plugins---fermyonactionsspinsetupv1) action instead?  (Nearly all written by rajatjindal 😄)

.github/workflows/test.yaml
karthik2804: We are just testing to `spin/npm` - particularly the `npm` packages we use in our build step.

.github/workflows/test.yaml
karthik2804: Will update.

vdice (approved):

karthik2804:
Will merge after I fix the failing `windows` test.

karthik2804:
I am thinking just building the app should be sufficient, do we also want to test running the app and getting a response? I do not have string preferences here.

vdice:
> I am thinking just building the app should be sufficient, do we also want to test running the app and getting a response? I do not have string preferences here.

What you've added already seems great for now.  Can be a follow-up if we end up wanting that coverage?

PR #282, opened 2024-08-26
Fix knitwit generation

karthik2804:
Moves the `knitwit` script to `postprepare` instead of `postinsall` which seems to lead to errors when cloning a repository that already has a `package-lock.json` but the modules are not installed.


karthik2804:
Happy to swap to a pre-release if we would like to use that to test the fix.

radu-matei (approved):

PR #286, opened 2024-08-29
add templates for redis

karthik2804:
This should probably be merged after #287 and bump the versions in the templates.

### DIFF
--- a/templates/http-js/content/package.json
+++ b/templates/http-js/content/package.json
@@ -18,6 +18,6 @@
         "@fermyon/knitwit": "https://github.com/fermyon/knitwit"
     },
     "dependencies": {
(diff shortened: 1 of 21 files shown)


karthik2804:
Are we okay with expecting a user to implement the class and export an object? If not, we could potentially add a `spin.ts` as we do for the `http` templates and wrap all that up where the user can just export one function.

radu-matei:
Given they're technically part of the same SDK, I'd like the two to be consistent.

karthik2804:
One part I would like to note though is that - the `http` handler uses the built in `fetchEvent` while redis and anything else would end up using generated bindings.

radu-matei (approved):

PR #287, opened 2024-08-29
bump version number to 2.1.0

karthik2804:
Version bump for a new minor release following the bump in the `ComponentizeJS` versions. This will allow us to add templates for the redis trigger as well.


tschneidereit (approved):
Would you mind also adding a tag for the release, and ideally also backfilling tags for the other 2.x releases so they show up in GitHub's UI?

karthik2804:
Will do @tschneidereit once I publish the npm release.

PR #288, opened 2024-09-06
Bump ComponentizeJS to optionally enable AoT

karthik2804:
This PR updates the following
- updates the ComponentizeJS dependency to bring in AoT using weval.
- Bumps the package version in preparation for a release.
- Updates the templates and test to be able to use AoT and the  newer package.

### DIFF
--- a/bin/j2w.mjs
+++ b/bin/j2w.mjs
@@ -30,6 +30,10 @@ const args = yargs(hideBin(process.argv))
         describe: "Spin trigger to target",
         demandOption: true
     })
+    .option('aot', {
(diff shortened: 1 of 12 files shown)


karthik2804:
Note: need to bump the version of spin_sdk dep in the template once a new release is done.

radu-matei (approved):
The one thing I'm not sure about is whether we want to enable this by default.

Given the impact on the application size, and that it's not immediately relevant for applications that are not compute intensive, I would be in favor of making this an opt-in feature, rather than opt-out.

PR #290, opened 2024-11-05
Bump componentizeJS to v0.14.0

radu-matei:


karthik2804:
@radu-matei I am going to close this as we cannot currently update to `componentizeJS 0.14.0` as that depends on `wasi@0.2.2` support which Spin does not support yet. I will create a follow-up PR to bump to `0.13.1`.

PR #296, opened 2024-11-21
make spin2 default and bump patch version

karthik2804:
Importing `spin3-imports` by default leads to `PostgresV3` imports making it into the binary because treeshaking is partially disabled. This fixes so that if users need to use Spin v3 imports they would have to opt in by specifying in the `knitwit.json`.

### DIFF
--- a/src/postgresv3.ts
+++ b/src/postgresv3.ts
@@ -1,3 +1,11 @@
+/**
+ * PostgresV3 enables interacting with a Postgres database using Spin's v3 interface.
+ * To use this module you  need to add `spin3-imports` to your `knitwit.json` file and run
+ * `npx knitwit`.
+ *
+ * @module PostgresV3
+ */
+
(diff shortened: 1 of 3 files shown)


src/postgresv3.ts
```diff
@@ -1,3 +1,11 @@
+/**
+ * PostgresV3 enables interacting with a Postgres database using Spin's v3 interface.
```
itowlson, line 2: You're doing this explicitly for Postgres 3.0 but not for other Spin 3 specific interfaces.  Is that intentional?  Does JS not import the wasi-cloud-core stuff?

karthik2804: The SDK does not explicitly implement bindings for the `wasi-cloud-core` stuff, meaning they will not be documented within the `spin-sdk` reference docs.

dicej (approved):

PR #305, opened 2025-01-15
[templates]: Add support for yarn

ThorstenHans:
This PR adds support for `yarn` as an alternative package manager to all templates in the repository.


templates/http-ts/metadata/snippets/component.txt
```diff
@@ -7,5 +7,5 @@ source = "{{ output-path }}/target/{{project-name | kebab_case}}.wasm"
 allowed_outbound_hosts = []
 
 [component.{{project-name | kebab_case}}.build]
-command = "npm run build"
+command = "{{ package-manager }} run build""
```
karthik2804:
```suggestion
command = "{{ package-manager }} run build"
```

There is an extra double quote in all the templates - this should fix CI

karthik2804:
@ThorstenHans the above suggested fix needs to be fixed for all the templates.

thorstenhans:
already pulled the branch... will update in a sec

radu-matei:
Anything else blocking this PR?

thorstenhans:
IMO, we wanted to time this PR with recent refactorings. Adding @karthik2804

thorstenhans:
I've rebased this PR and incorporated all the recent changes. cc @karthik2804

karthik2804 (approved):

thorstenhans:
We're currently facing an issue when using `yarn` as package manager. Although, building the Spin component does not result in an error, the resulting component won't run at all. 

The wasm component does not provide necessary exports. It looks like knitwit does not load `KNITWIT_SOURCE` from the corresponding environment variable. Instead, it's only pulling from `npm_config_knitwit_source` or falling back to defaults... cc @karthik2804

karthik2804:
The issue seems to be how `npm` and `yarn` treat case sensitivity differently for env vars in `.npmrc`. Marking as draft as in the next major release, we will not need to depend on the env vars for build and add `yarn` as a part of it.

karthik2804:
Not planned for now.

PR #306, opened 2025-01-20
upgrade knitwit and add caching to build script

karthik2804:
We can upgrade the Componentize-js once there is a new release. This PR will allow us to create a new release which can then be used to restructure templates.

This PR also adds caching to avoid componentization during the build if not required.

### DIFF
--- a/bin/j2w.mjs
+++ b/bin/j2w.mjs
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
 import { componentize } from "@bytecodealliance/componentize-js";
-import { readFile, writeFile } from 'node:fs/promises';
+import { readFile, writeFile, access } from 'node:fs/promises';
+import { createHash } from 'node:crypto';
 import { resolve, basename } from 'node:path';
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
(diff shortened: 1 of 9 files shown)


test/test-app/package.json
```diff
@@ -4,9 +4,9 @@
     "description": "",
     "main": "index.js",
     "scripts": {
-        "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i dist.js -d combined-wit -n combined -o target/test-app.wasm --aot",
+        "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i build/bundle.js -d build/wit/knitwit -n combined -o target/test-app.wasm --aot",
```
thorstenhans: IIRC we wanted to rename the final output folder, to have something different than `target`

bin/j2w.mjs
```diff
@@ -37,23 +38,68 @@ const args = yargs(hideBin(process.argv))
...
+
+async function saveChecksum(checksumPath, checksum) {
+    await writeFile(checksumPath, checksum);
+}
+
+(async () => {
+    const sourceChecksum = await calculateChecksum(src);
+    const existingChecksum = await getExistingChecksum(inputChecksumPath);
+
+    if ((existingChecksum === sourceChecksum) && fileExists(outputPath)) {
```
thorstenhans, line 77: IMO we should also create a checksum for `package-lock.json`. Shouldn't we? User could update dependencies or dev-dependencies that could affect the build.

thorstenhans (commented):
Raised two questions as part of my review

bin/j2w.mjs
karthik2804: That will have an impact on the bundle generated by webpack which is what would be the input to this build script. We still end up running webpack everytime.

test/test-app/package.json
karthik2804: Ah yea `dist`, I will rename.

radu-matei (approved):

PR #307, opened 2025-01-20
Restructure templates

karthik2804:
This PR restructures the templates so that the default directory structure follows more javascript-ish structure and tucks away knitwit as the user should not have to deal with it usually. It also updates the default HTTP templates to now have a router with a catch-all 404 set up.


templates/http-js/content/src/index.js
```diff
@@ -1,6 +1,18 @@
-import { ResponseBuilder } from "@fermyon/spin-sdk";
+import { ResponseBuilder, Router } from "@fermyon/spin-sdk";
 
-export async function handler(req, res) {
-    console.log(req);
+let router = Router();
+router.get("*", (_, req, res) => { handleDefaultRoute(req, res) })
```
radu-matei: potential nit that you can ignore: instead of separate functions, lambdas here?

templates/http-js/content/src/index.js
```diff
@@ -1,6 +1,18 @@
...
 }
+
+async function notFound(req, res) {
+    res.status(404);
+    res.send("not found");
+}
+
+export async function handler(req, res) {
+    await router.handleRequest(req, res);
+}
```
tschneidereit: Missing newline at the end
```suggestion
}

```

templates/http-js/content/src/index.js
```diff
@@ -1,6 +1,18 @@
...
 
-export async function handler(req, res) {
-    console.log(req);
+let router = Router();
+router.get("*", (_, req, res) => { handleDefaultRoute(req, res) })
+router.all("*", (_, req, res) => { notFound(req, res) })
+
+async function handleDefaultRoute(req, res) {
     res.send("hello universe");
 }
```
tschneidereit: I think it'd be good to give the routes a bit more structure, and to add some comments here to help people understand where they should make which edits:
```suggestion
let router = Router();
// Modify this route or add additional ones to implement the component's API:
router.get("/hello/:name", (metadata, req, res) => { handleHelloRoute(req, res, metadata.params) });

// Default route that will be called for any routes not handled above:
router.all("*", (_, req, res) => { notFound(req, res) });

async function handleHelloRoute(req, res, { name }) {
    res.send(`hello ${name}`);
}
```

templates/http-ts/content/src/index.ts
```diff
@@ -1,6 +1,18 @@
...
 }
+
+async function notFound(req: Request, res: ResponseBuilder) {
+    res.status(404);
+    res.send("not found");
+}
+
+export async function handler(req: Request, res: ResponseBuilder) {
+    await router.handleRequest(req, res);
+}
```
tschneidereit:
```suggestion
}

```

templates/http-ts/content/src/index.ts
```diff
@@ -1,6 +1,18 @@
-import { ResponseBuilder } from "@fermyon/spin-sdk";
+import { ResponseBuilder, Router } from "@fermyon/spin-sdk";
 
-export async function handler(req: Request, res: ResponseBuilder) {
-    console.log(req);
+let router = Router();
```
tschneidereit, line 3: Would be good to apply the changes from the JS version to this version, too.

tschneidereit (approved):
Looks good to me! I added a few suggestions, but nothing major

templates/http-js/content/src/index.js
karthik2804: I have a slight preference  in this case to use functions as it keeps the router structure easily readable when more are added.

karthik2804:
@tschneidereit updated to be in line with the suggestions.

tschneidereit (approved):
Thank you!

PR #308, opened 2025-01-22
bump componentizeJS and make build script dependant on the version of CJS as well

karthik2804:
This PR bumps ComponentizeJS which required us to update the `wit` files. This update also bring in a ton of new stuff like an updated StarlingMonkey engineer that supports blobs and formdata and also bug fixes which enable the use of AWS SDK again.

The build script is also updated to take into consideration the ComponentizeJs version that was used to build now.


karthik2804:
superseded by #317

PR #309, opened 2025-01-22
unbreak spin CI test

karthik2804:
Adds a default route to unbreak CI tests for Spin.


itowlson (approved):
Thanks!

(also, for what it's worth, I think this does more than "unbreak CI tests for Spin" - it means that when someone does a `spin up`, and curls or clicks the link that `spin up` displays, they see success rather than a 404 - so very glad to see this from that point of view too!)

PR #317, opened 2025-01-24
Prepare for v3.0.0-rc1

karthik2804:
* Bump componentizeJS and make build script dependant on the version of CJS as well
* Drop responseProvider and bump router to v5
* Update test to add streaming test and use new router

The test currently serves as an example application. Once we have a new release, we need to update the templates and examples.

**TODO**:
- [ ] Move the `knitwit` step in test to build instead of postinstall (this needs to be done) for the templates as well)


radu-matei (approved):
Unblocking so we can try the RC out.

Thanks!

PR #326, opened 2025-02-04
add precompilation for regex during the build step

karthik2804:
This PR utilizes https://github.com/fastly/js-compute-runtime/blob/main/src/precompile.js to precompile regex during wizer time. This enables compatibility with more libraries.

The code imported from Fastly retains its own license header to ensure attribution of authorship.

This PR also bumps the version for the next release.


tschneidereit (approved):
Looks great! I'll edit the commit message before landing this to call out very explicitly that this pulls in Fastly code, to give proper attribution

PR #333, opened 2025-02-11
[examples]: Updated router-sample

thorstenhans:
This PR updates the router sample to illustrate how users could define different types of handlers. Also, a `POST` endpoint is registered for demonstration purposes.

### DIFF
--- a/examples/common-patterns/routing-requests/src/index.ts
+++ b/examples/common-patterns/routing-requests/src/index.ts
@@ -11,7 +11,7 @@ router
   .get('/hello/:name', ({ name }) => `Hello, ${name}!`)
   .get('/bye/:name', ({ name }) => sayGoodBye(name))
   .get('/reverse-header-value', req => reverseHeaderValue(req))
-  .post('/items', async (req) => storeItem(await req.arrayBuffer()));
+  .post('/items', async (req) => processItem(await req.arrayBuffer()));
 
 
 //@ts-ignore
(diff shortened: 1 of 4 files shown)


karthik2804 (commented):
overall LGTM- I think we should probably update the readme with the curl commands to make the usage clear.

thorstenhans:
@karthik2804 I rebased the PR and updated the README for the router sample 👀

examples/common-patterns/routing-requests/src/index.ts
```diff
@@ -7,10 +7,51 @@ let router = AutoRouter();
...
+  let payload;
+  try {
+    payload = JSON.parse(decoder.decode(requestBody));
+  }
+  catch (error) {
+    return new Response('Bad Request', { status: 400 });
+  }
+  console.log(`Storing item ${JSON.stringify(payload)}`);
+  return new Response('Item stored', { status: 201 });
+};
```
karthik2804: We probably want to leave a comment that it is not actually storing things. Something like a comment block after parsing the payload might be a good idea.

karthik2804 (approved):
LGTM - one comment which is not a blocker

examples/common-patterns/routing-requests/src/index.ts
thorstenhans: Decided to rename the handler and corresponding log messages with `processing item` instead.

radu-matei (approved):

Issue #324, opened 2025-02-03
Release CI


radu-matei:
We need to update our release process to publish the NPM package from CI.

thorstenhans:
This should also include generating the docs. Once #335 is merged, we could simply invoke `npm run build-docs`, the version plugin for typedoc will take version from `package.json` into context.

PR #334, opened 2025-02-11
add action to automate publish releases to npm

karthik2804:
Adds action that publishes package to npm when a tag starting with `v` is pushed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,6 +22,18 @@ jobs:
           node-version: 22
           registry-url: 'https://registry.npmjs.org/'
 
+      - name: Verify tag matches package version
+        run: |
+          PACKAGE_VERSION=$(jq -r .version package.json)
+          TAG_VERSION=${{ github.ref_name }}
+          # Remove 'v' prefix from the tag if it exists
+          TAG_VERSION=${TAG_VERSION#v}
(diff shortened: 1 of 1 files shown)


thorstenhans:
@karthik2804 see my comment in #324 about generating docs

karthik2804:
@ThorstenHans I like the idea of releasing docs per release but would prefer not to roll that into this PR as that will involve us having to generate and commit docs or push it to a package. I am happy to tackle that as a follow-up.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build the project
+        run: npm run build
+
+        # Determine if the tag is a release candidate or not
+      - name: Determine npm tag
+        id: npm_tag
```
radu-matei, line 45: nit: why the ID for this job but not for anything else?

radu-matei (approved):

.github/workflows/release.yaml
karthik2804: It can be safely removed.

karthik2804:
Closing this as we will need to work on a different action once we have v4 because each interface becomes modularized.

PR #356, opened 2025-03-19
Add support for hono, itty and no router to `http-ts` and `http-js`

thorstenhans:
## UPDATE

Instead of adding new templates, this PR modifies existing `http-ts` and `http-js` templates. Users could now select if they wanna use `itty` or `hono`. They could also go without a router by responding to the particular prompt with `none`.

## Original 

This PR adds the two new templates to the Spin SDK for JavaScript:

- `http-hono-ts`: HTTP triggered Spin application using TypeScript and Hono
- `http-hono-js`: HTTP triggered Spin application using JavaScript and Hono


radu-matei (approved):
Tested the TS template, LGTM!

Thanks!

thorstenhans:
Q: @radu-matei @karthik2804 do we wanna use the suggested names? (`http-hono-ts` & `http-hono-js`)? Mainly asking because Karthik suggested `http-ts-hono` and `http-js-hono`)

radu-matei:
`http-*s-hono` seems more consistent with the `http-<language>` format.

karthik2804:
I prefer the consistency as well. One potential alternate for us to consider is if we want to consider just having a single template that prompts for choices?  similar to the work on the package manager PR we have open. The advantage would be that we have one single template but the flipside is the user has to make more choices and that the template will look a lot more bloated with if statements( It is not something that end users will have to deal with though)

itowlson:
FWIW I've previously used the language-last convention (e.g. https://github.com/itowlson/spin-templates-slack/tree/main/templates/slack-slash-command-rust).  But that is definitely not any kind of established project precedent.

@karthik2804 Yes, it would be even better to have a template that asks users "which router do you want to use (itty / hono / none)".  It would definitely make the template gnarly though (and I think we'd need to add a choice type to the system, I don't think we have that yet).

karthik2804:
We can make choices based on string matching already IIRC.

thorstenhans:
How should we proceed here? I'm happy to rename the templates to `http-{lang}-hono`

If we extend prompting for the default http-js / http-ts templates, I would also vote for adding a prompt to chose the desired package manager (npm|yarn|pnpm).

Any ideas on how to proceed here in the meantime?

karthik2804:
>I would also vote for adding a prompt to chose the desired package manager (npm|yarn|pnpm).
@ThorstenHans as we have found out `yarn` does not work with the v3 build tooling, so that will have to wait until v4 but I agree that it is a good idea to add it as well once that is a possibility.

I am happy to go eitherway on the template but have a slight preference on the "extending prompting" approach as that will mean lesser templates.

thorstenhans:
I've updated the templates `http-js` and `http-ts` now. Until the PR mentioned by @itowlson lands in spin:main, I added a question with a regex that allows the user to chose between `hono`, `itty`, or go without a router.

thorstenhans:
@karthik2804 I updated both templates `http-ts` and `http-js` as we discussed. Also the matrix strategy for testing templates has been updated

karthik2804 (approved):
This looks good to me!
I am just curious if we are able to use partials and split things into multiple files?

itowlson:
@karthik2804 Looking at how Liquid partials work (https://docs.rs/liquid/latest/liquid/partials/index.html)... the template system doesn't currently do any of that.  No reason it couldn't, but we'd have to do some work on it, which might be fairly easy, but would still mean not until Spin 3.3.  Sorry - I can see the value, just hasn't been needed before.

karthik2804 (approved):

PR #352, opened 2025-03-10
The massive migration to the v4 SDK

karthik2804:
This PR brings in a bunch of changes as we work towards V4 of the SDK.
- It modularizes the SDK into multiple packages (one per wasi interface). This allows consumers to import only the required interfaces and thus have the output component rely only on the used interfaces instead of all without having to rely on tree shaking. 
- It introduces a new `build-tools` package that contains an updated build script that takes over the management of the application `wit` imports and exports, eliminating the need for `knitwit`. The build system now recursively checks for the dependencies in the `package.json` and then uses that to generate the target world wit using which the source is componentized.
- As a dependency of the build-tools, we have updated the packages to follow a new convention for configuring the `wit` dependencies on the `package.json`

**NOTE**: The test app can be considered as an example of how things can be expected to look. 

TODOS:
- [x] figure out versioning for the individual packages. 
- [ ] update the examples and templates once the various packages are released.
- [ ] figure out CI for releases of the various packages.

fixes #345, fixes #346

PR #357, opened 2025-03-26
V4 template update

karthik2804:
I will rebase on #352 once that is merged.

merge after #352


thorstenhans (approved):

radu-matei (approved):
Let's goo!

LGTM

templates/http-js/content/package.json
```diff
@@ -13,11 +13,11 @@
...
         "mkdirp": "^3.0.1",
         "webpack": "^5.74.0",
-        "webpack-cli": "^4.10.0",
-        "@fermyon/knitwit": "0.3.0"
+        "webpack-cli": "^4.10.0"
     },
     "dependencies": {
-        "@fermyon/spin-sdk": "^3.0.0",
+        "@spinframework/build-tools": "^1.0.1",
+        "@spinframework/wasi-http-proxy": "1.0.0",
```
radu-matei: Not including the other packages here means discoverability is quite hard.

I _think_ we should include them?

What do you think @tschneidereit ?

radu-matei: Why pin one to a specific version and use `^` for the other?

radu-matei:
nit (not changed in this PR): inconsistent quotes across the source code:

```
    .get("/", () => new Response("hello universe"))
    .get('/hello/:name', ({ name }) => `Hello, ${name}!`)
```

templates/http-js/content/package.json
karthik2804: It is just inconsistency, we should add the `^`

karthik2804: This was a result of the conversation @tschneidereit and I had. We wanted to start from a minimal subset so that by default we build apps that run in different runtimes. We now have a readme that lists the packages one can use.

karthik2804: fixed

karthik2804:
Fixed the inconsistent quotes.

PR #359, opened 2025-04-07
(packages/build-tools): Bump componentizeJS and fix windows paths

karthik2804:
This PR brings in fixes for windows paths when trying build with the tool. It converts the windows paths to ones that can be used in wasi. This PR also bumps the version of the package.

### DIFF
--- a/packages/build-tools/src/wasiDepsParser.ts
+++ b/packages/build-tools/src/wasiDepsParser.ts
@@ -1,6 +1,24 @@
 import fs from 'fs';
 import { TargetWorld } from '../lib/wit_tools.js';
 import path from 'path';
+import { resolve } from 'node:path';
+import { platform } from 'node:process';
+const isWindows = platform === 'win32';
+
+function maybeWindowsPath(path: string): string {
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/wasiDepsParser.ts
```diff
@@ -1,6 +1,24 @@
 import fs from 'fs';
 import { TargetWorld } from '../lib/wit_tools.js';
 import path from 'path';
+import { resolve } from 'node:path';
+import { platform } from 'node:process';
+const isWindows = platform === 'win32';
+
```
tschneidereit, line 7: I don't think this is technically copyrightable, but we should still give attribution. Luckily the license is the same, so a simple link is sufficient.
```suggestion

// Copied from https://github.com/bytecodealliance/ComponentizeJS/blob/e5bb0da0c1ce1ee87eed45c970121ed5fe8cca3a/src/componentize.js#L26-L39
```

tschneidereit (approved):
LGTM!

PR #360, opened 2025-04-15
Generate docs for modules

karthik2804:
This PR does a couple of things
- Removes the old docs generation script and only generates the current version
- Introduces a package.json at the root level as the first step towards converting this to a npm workspace. This is currently used to generate docs.

Note: Once a release/manual dispatch occurs this will remove the docs of the older versions of the SDK away. We will make a copy of the docs and publish it separately otherwise there will be versioning confusion.


radu-matei (approved):

package.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "spin-js-sdk",
+    "version": "1.0.0",
+    "description": "This is an SDK for Javascript and Typescript based on [ComponentizeJS](https://github.com/bytecodealliance/ComponentizeJS).",
```
tschneidereit: Should this description reflect that this is just for the docs?

tschneidereit (approved):

PR #361, opened 2025-04-15
add regex precompilation to build step

karthik2804:
alternatively, we could write the precompiled source to a file and keep the new syntax. @tschneidereit Do you have a preference.

### DIFF
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -65,7 +65,7 @@ async function main() {
     const source = await readFile(src, 'utf8');
     const precompiledSource = precompile(source, src, true) as string;
 
-    // Using the old syntax because the new sytnax does not allow passing in the precompiled string
+    // Using the old syntax because the new syntax does not allow passing in the precompiled string
     //@ts-ignore
     const { component } = await componentize(precompiledSource, inlineWit, {
       runtimeArgs,
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -13,6 +13,7 @@ import { readFile, writeFile } from 'node:fs/promises';
 import { mergeWit } from '../lib/wit_tools.js';
 //@ts-ignore
 import { precompile } from "./precompile.js"
+import path from 'node:path'
 
 async function main() {
   try {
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/index.ts
```diff
@@ -61,10 +62,12 @@ async function main() {
...
     );
 
-    const { component } = await componentize({
-      sourcePath: src,
-      // @ts-ignore
-      witWorld: inlineWit,
+    const source = await readFile(src, 'utf8');
+    const precompiledSource = precompile(source, src, true) as string;
+
+    // Using the old syntax because the new sytnax does not allow passing in the precompiled string
```
tschneidereit:
```suggestion
    // Using the old syntax because the new syntax does not allow passing in the precompiled string
```

test/test.sh
```diff
@@ -36,7 +35,7 @@ echo "\n\nTest completed"
 
 # kill the spin app
 echo "Stopping Spin"
-kill -9 $SPIN_PID
+killall spin
```
tschneidereit, line 38: I think you can keep this to the specific process by changing from `kill -9` to `kill -6`. Can you test that and see if it works? I had this result in very unexpected behavior before, and am certain that we (and others) would run into it again in the future.

tschneidereit (approved):
The problem with this approach is that it will make debugging impossible, unfortunately. I think it's okay to land for now, but we'll have to restructure it pretty substantially before too long.

What I imagine we'll end up doing is along the lines of
- instead of relying on bundles, support walking all imports and looking for regexs in all of them
- then, instead of overwriting regex occurrences in the source, we collect them and write compilation steps for them into a new file
- finally, we pass a new file as input to componentize that imports the original main file and this generated file

I think that should make everything work, keep source maps intact, etc.

But for now, let's land this so we get the speedup :)

karthik2804:
@tschneidereit Do we want to optionally enable the regex stuff, so that if debugging is enabled, we can ignore the precompile step for the time being? Is that an acceptable tradeoff?

tschneidereit:
oh, that's a great idea! Let's maybe tie it to the `-d` flag and disable the precompilation automatically if that flag is set?

karthik2804:
The only downside is that we will not be able to compile applications that have unicode regex in debug, but I am okay with the tradeoff.
An alternative I see is that when debugging, we can write the output to a file instead of in memory, so that you can still have stack traces to something on the disk.

tschneidereit:
Yeah, we could do that, and it might be nicer for the time being

test/test.sh
karthik2804: This did not work, I can take a look at how to fix this in a separate PR.

PR #371, opened 2025-07-18
Fix build caching

karthik2804:
Add a field to track the target world for which the build is done so that we re componentize when target world changes but the source does not.

### DIFF
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -4,7 +4,8 @@ import { componentize } from '@bytecodealliance/componentize-js';
 import { version as componentizeVersion } from '@bytecodealliance/componentize-js';
 import { getPackagesWithWasiDeps, processWasiDeps } from './wasiDepsParser.js';
 import {
-  calculateChecksum,
+  calculateCheckSum,
+  calculateFileChecksum,
   saveBuildData,
 } from './utils.js';
 import { getCliArgs } from './cli.js';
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/index.ts
```diff
@@ -63,6 +55,17 @@ async function main() {
       'combined-wit:combined-wit@0.3.0',
     );
 
+    // Calcuilate the checksum of the inline wit
```
tschneidereit:
```suggestion
    // Calculate the checksum of the inline wit
```

tschneidereit: Though I would actually favor just removing this comment: it says exactly what the code says, after all

packages/build-tools/src/utils.ts
```diff
@@ -4,7 +4,7 @@ import { access, writeFile } from 'node:fs/promises';
 
 
 // Function to calculate file checksum
-export async function calculateChecksum(filePath: string) {
+export async function calculateFileChecksum(filePath: string) {
   try {
     const fileBuffer = await readFile(filePath);
     const hash = createHash('sha256');
```
tschneidereit, line 9: (I can't suggest changes spanning unmodified lines, sorry)

Replace lines 10-12 with `return calculateChecksum(fileBuffer)` to reduce duplication and ensure that we always use the same way of hashing things.

packages/build-tools/src/utils.ts
```diff
@@ -16,6 +16,12 @@ export async function calculateChecksum(filePath: string) {
   }
 }
 
+export function calculateCheckSum(data: string | Buffer): string {
```
tschneidereit: Let's keep the spelling the same for both functions:
```suggestion
export function calculateChecksum(data: string | Buffer): string {
```

tschneidereit (approved):

PR #372, opened 2025-07-22
add ability to generate chained sourcemap for precompiled sources

karthik2804:
This PR adds the capability to the build tooling to chain the source map for the precompiled source with the input source map, allowing for better debugging.

In a follow-up PR, the plan is to move to build instead of webpack because the source maps generated by build are friendlier.

cc: @itowlson


karthik2804:
Converting to draft to fix up a couple of alignment issues.

karthik2804:
The test has been updated to use esbuild and now the `precompiled-source.js.map` which is generated along with `precompiled-source.js` map back directly to the sources chaining the one generated by the build process. Note that we do not resolve maps inside the `node_modules` folder as some NPM packages have sourcemaps but not the source bundled.

karthik2804:
@tschneidereit friendly bump for a review. Thanks!

packages/build-tools/src/precompile.js
```diff
@@ -51,12 +51,23 @@ export function precompile(source, filename = '<input>', moduleMode = false) {
...
+    let match;
+    while ((match = sourceMapRegex.exec(source))) {
+        const start = match.index;
+        const end = start + match[0].length;
+        magicString.remove(start, end);
+    }
+
+    const precompiledSource = magicString.toString() + `\n//# sourceMappingURL=${precompiledFileName}.map\n`;
+
     // When we're ready to pipe in source maps:
```
tschneidereit:
```suggestion
```

test/test-app/build.mjs
```diff
@@ -0,0 +1,39 @@
...
+let SourceMapPlugin = {
+    name: 'excludeVendorFromSourceMap',
+    setup(build) {
+        build.onLoad({ filter: /node_modules/ }, args => {
+            return {
+                contents: fs.readFileSync(args.path, 'utf8')
+                    + '\n//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIiJdLCJtYXBwaW5ncyI6IkEifQ==',
+                loader: 'default',
+            }
+        })
```
tschneidereit, line 22: Not blocking anything, but it'd be nice to have a comment here explaining how this test works: I have no idea how it does, myself :)

tschneidereit (approved):
Apologies for the delay!

test/test-app/build.mjs
karthik2804: Updated with a comment. But this basically stubs sourcemaps references by npm packages to avoid issues where the sourcemap refer to files not included in the package.

PR #373, opened 2025-08-12
Debugger test app

itowlson:
As part of the StarlingMonkey debugger work (https://github.com/bytecodealliance/StarlingMonkey/pull/248) I wanted to test against the Spin runtime as well as the Wasmtime CLI.  However it was felt that we should put tests that ran in runtimes other than Wasmtime into the repos associated with those runtimes rather than including them with the other debugger tests.

At the moment this is the only test app I have.  @karthik2804 mentioned that he had demonstrated debugging working with a bundler (esbuild? not sure), so we should add a test case for that too.

### DIFF
--- a/test/debugger-testing/spin-ts/README.md
+++ b/test/debugger-testing/spin-ts/README.md
@@ -1,0 +1,32 @@
+# `http-ts` Template
+
+A starter template for building TypeScript HTTP applications with Spin.
+
+## Getting Started
+
+Build the App
+
+```bash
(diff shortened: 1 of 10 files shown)


test/debugger-testing/spin-ts/README.md
karthik2804, line 1: Can we update this readme to reflect how one would use the debugger?

karthik2804 (approved):

PR #378, opened 2025-09-12
build_tool: update cjs to v0.19.0 and bump package version

karthik2804:
Bumps ComponentizeJS  which brings in a newer build of StarlingMonkey but we loose `aot`. I have currently not removed the option but just print a warning and continue with building without it. If there are strong preferences, I am happy to delete it. 

It will be a breaking change, strictly speaking, but I do not think `aot` has been advertised in the Spin docs, and I'm also not sure if it is widely used.

### DIFF
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -85,6 +85,14 @@ async function main() {
       await writeFile(precompiledSourcePath + '.map', JSON.stringify(precompiledSourceMap, null, 2));
     }
 
+    // if aot is enabled, warn that it has been temporarily disabled
+    if (CliArgs.aot) {
+      console.warn(
+        'Warning: AOT compilation is temporarily disabled due to issues with componentize-js. Proceeding without AOT. It may be removed in future releases.',
+      );
+      CliArgs.aot = false;
+    }
+
     const { component } = await componentize({
       sourcePath: precompiledSourcePath,
       // @ts-ignore
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -87,10 +87,9 @@ async function main() {
 
     // if aot is enabled, warn that it has been temporarily disabled
     if (CliArgs.aot) {
-      console.warn(
-        'Warning: AOT compilation is temporarily disabled due to issues with componentize-js. Proceeding without AOT. It may be removed in future releases.',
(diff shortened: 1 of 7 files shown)


tschneidereit:
> Bumps ComponentizeJS which brings in a newer build of StarlingMonkey but we loose `aot`. I have currently not removed the option but just print a warning and continue with building without it. If there are strong preferences, I am happy to delete it.

I think the best thing for now would be to make it a hard error instead of a warning: we shouldn't just silently build something with very different properties from what the user expected.

> It will be a breaking change, strictly speaking, but I do not think `aot` has been advertised in the Spin docs, and I'm also not sure if it is widely used.

Given that we've never documented this, I think I agree that it's okay not to treat this as a breaking change, yeah. Over in CJS, we decided that it needs to be a breaking change, but there it was fully documented, so the situation was different.

packages/build-tools/src/index.ts
```diff
@@ -85,6 +85,14 @@ async function main() {
       await writeFile(precompiledSourcePath + '.map', JSON.stringify(precompiledSourceMap, null, 2));
     }
 
+    // if aot is enabled, warn that it has been temporarily disabled
+    if (CliArgs.aot) {
+      console.warn(
+        'Warning: AOT compilation is temporarily disabled due to issues with componentize-js. Proceeding without AOT. It may be removed in future releases.',
+      );
+      CliArgs.aot = false;
```
tschneidereit: As mentioned, I think we should make this a hard error, not a warning.
```suggestion
      throw new Error(
        'AOT compilation is currently unavailable. Remove the `aot` option to proceed.',
      );
```

Additionally, I think we should add a `hidden: true` field to the yargs options for `aot` over in `cli.ts`.

tschneidereit (approved):
LGTM with comment addressed.

packages/build-tools/src/index.ts
karthik2804: Between this option and potentially just removing the option. Do you have a preference?

tschneidereit: I think this option would be better: we don't know if someone looked at the `--help` output or the source and is using this in the wild after all. In that case, getting an explanation is definitely better than the option just disappearing.

karthik2804: Done will update.

PR #379, opened 2025-09-30
fix handling of KV set and throw wrapped errors

karthik2804:
Fixes a couple of small bugs that I ran into.

- We were not handling the case `ArrayBuffer` properly in KV set operations.
- The responses returned by starlingMonkey were cyclic, and routers like Itty threw errors when trying to go from an error to a response. This fix now extracts the useful parts of the error and wraps them up in a simpler JS error.

### DIFF
--- a/packages/spin-kv/src/index.ts
+++ b/packages/spin-kv/src/index.ts
@@ -59,6 +59,8 @@ function createKvStore(store: spinKv.Store): Store {
       if (!(value instanceof Uint8Array)) {
         if (typeof value === 'string') {
           value = encoder.encode(value);
+        } else if (value instanceof ArrayBuffer) {
+          value = new Uint8Array(value);
         } else if (typeof value === 'object') {
           value = encoder.encode(JSON.stringify(value));
         }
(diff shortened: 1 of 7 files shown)


karthik2804:
@tschneidereit I am not really sure about the error part and wonder if handling the error should be a user concern instead of us wrapping it.

packages/spin-kv/src/index.ts
```diff
@@ -59,6 +59,8 @@ function createKvStore(store: spinKv.Store): Store {
       if (!(value instanceof Uint8Array)) {
         if (typeof value === 'string') {
           value = encoder.encode(value);
+        } else if (value instanceof ArrayBuffer) {
+          value = new Uint8Array(value);
```
tschneidereit, line 63: Tiniest of nits: I would move this to the top so the `encode` cases stay grouped

tschneidereit (approved):
I think that's an okay workaround, yes. It's somewhat unfortunate that the generated bindings have cyclic structures, but we should look into tackling that with the move to wit-dylib in time.

radu-matei (approved):
Anything else before merging this?

Thanks!

PR #381, opened 2025-10-20
Update templates

karthik2804:
I need to verify that the debugger works.


radu-matei (commented):
This requires a rebase now to resolve conflicts.


Thanks!

templates/http-js/content/.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "starlingmonkey": {
+        "componentRuntime": {
+            "executable": "spin",
+            "options": [
+                "up",
```
tschneidereit, line 6: one thing to consider here is to change this to `watch`. That'd probably require/enable other changes to the lifecycle, as the process wouldn't need to be restarted when content changes.

templates/http-ts/content/.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "starlingmonkey": {
+        "componentRuntime": {
+            "executable": "spin",
+            "options": [
+                "up",
```
tschneidereit, line 6: Same comment as for the `http-js` template

test/test-app/.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "starlingmonkey": {
+        "componentRuntime": {
+            "executable": "spin",
+            "options": [
+                "up",
```
tschneidereit, line 6: ... and same here

tschneidereit (approved):
Looks great, thank you!

The only comments I left inline are about whether to switch from `spin up` to `spin watch`, which might resolve some issues with the debugger, but for all I know might also introduce others. If you haven't already, I think testing that would be good.

karthik2804:
@tschneidereit Thoughts on landing this as is and I can test the `spin watch` change and open a future PR?

karthik2804:
I also think we would need build profiles to tightly integrate with the `spin build` command.

tschneidereit:
> I also think we would need build profiles to tightly integrate with the `spin build` command.

Oh, that's a good point! Yeah, with that in mind I agree that landing this as-is makes sense.

tschneidereit:
This makes me realize that it'd be valuable to have a `watch` mode that doesn't trigger a rebuild, but just a reload of the wasm file if that changes

itowlson:
> This makes me realize that it'd be valuable to have a `watch` mode that doesn't trigger a rebuild, but just a reload of the wasm file if that changes

`spin watch --skip-build` unless I'm misunderstanding you

tschneidereit:
You're not: I had forgotten that exists! @karthik2804 might be useful to update the templates to use that, since it'd allow us to avoid having to restart `spin` between debugging sessions

itowlson:
I have a feeling that `spin watch` starts new `spin up/spin trigger http` processes so I am not sure that any debugger attachment will remain, but I'm not sufficiently confident with either `spin watch` _or_ the debugger to say for sure...!

PR #382, opened 2025-10-20
update componentizeJS to v0.19.2

karthik2804:
Pull in the latest componentizeJS and add a `initLocation` flag to use the newly added ability in StarlingMonkey to have location enabled during top level init.

### DIFF
--- a/packages/build-tools/src/cli.ts
+++ b/packages/build-tools/src/cli.ts
@@ -6,6 +6,7 @@ export interface CliArgs {
   output: string;
   aot?: boolean;
   debug?: boolean;
+  initLocation?: string;
 }
 
 export function getCliArgs(): CliArgs {
@@ -29,6 +30,10 @@ export function getCliArgs(): CliArgs {
       alias: 'd',
       describe: 'Enable JavaScript debugging',
       type: 'boolean',
+    })
+    .option('initLocation', {
+      describe: 'Location during top level initialization',
+      type: 'string',
     })
     .argv as CliArgs;
 
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -26,7 +26,10 @@ async function main() {
     let CliArgs = getCliArgs();
     let src = CliArgs.input;
     let outputPath = CliArgs.output;
-    let runtimeArgs = CliArgs.debug ? '--enable-script-debugging' : '';
+    let runtimeArgs = [
+      CliArgs.debug && '--enable-script-debugging',
+      CliArgs.initLocation && `--init-location ${CliArgs.initLocation}`
+    ].filter(Boolean).join(' ');
 
     // generate wit world string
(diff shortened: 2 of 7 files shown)


packages/build-tools/src/cli.ts
```diff
@@ -30,6 +31,10 @@ export function getCliArgs(): CliArgs {
       describe: 'Enable JavaScript debugging',
       type: 'boolean',
     })
+    .option('initLocation', {
+      describe: 'Location during top level initialization',
```
tschneidereit, line 35: I think a slightly more explicit description would be good here:
```suggestion
      describe: 'URL used for the "Location" builtin during top level initialization',
```

packages/build-tools/src/index.ts
```diff
@@ -26,7 +26,10 @@ async function main() {
     let CliArgs = getCliArgs();
     let src = CliArgs.input;
     let outputPath = CliArgs.output;
-    let runtimeArgs = CliArgs.debug ? '--enable-script-debugging' : '';
+    let runtimeArgs = [
+      CliArgs.debug && '--enable-script-debugging',
+      CliArgs.initLocation && `--init-location ${CliArgs.initLocation}`
+    ].filter(Boolean).join(' ');
```
tschneidereit, line 32: This is too clever for my taste (where my taste is informed by the fact that I needed to really analyze how this works to assure myself that it does what it should).

How about this instead? Slightly cleverer than `if` statements and a bit less verbose, but making clearer (to me, at least) what's going on:
```suggestion
    let runtimeArgs = [];
    CliArgs.debug && runtimeArgs.push('--enable-script-debugging');
    CliArgs.initLocation && runtimeArgs.push(`--init-location ${CliArgs.initLocation}`);
```

test/test-app/src/index.ts
```diff
@@ -1,6 +1,13 @@
 import { AutoRouter } from "itty-router"
 import { headersTest, health, kvTest, kvTestUint8Array, outboundHttp, statusTest, stream, streamTest, testFunctionality } from "./test";
 
+// Verify top level init location
+if (globalThis.location.href != "http://foo.bar/") {
+    throw new Error(`Expected top-level init location to be "top-level", got "${globalThis.location}"`);
```
tschneidereit:
```suggestion
    throw new Error(`Expected top-level init location to be "http://foo.bar/", got "${globalThis.location}"`);
```

test/test-app/src/index.ts
```diff
@@ -1,6 +1,13 @@
...
+} else {
+    console.log("Top-level init location verified:", globalThis.location.href);
```
tschneidereit: Might as well be consistent with the use of strings vs templates:
```suggestion
    console.log(`Top-level init location verified: ${globalThis.location.href}`);
```

tschneidereit (approved):
Looks great!

PR #384, opened 2025-10-27
set default localtion during top level init in templates

karthik2804:


thorstenhans:
LGTM - I tested with two different apps and can confirm that it works as expected

thorstenhans (approved):
🚢

PR #385, opened 2025-11-04
Fix precompile

karthik2804:
This PR fixes an issue discovered by @kate-goldenring. The default template without a router failed to compile because it did not have any regex, and the return of `precompile` was a string rather than an object. This PR gives the `precompile` function a consistent return type.

### DIFF
--- a/packages/build-tools/package-lock.json
+++ b/packages/build-tools/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@spinframework/build-tools",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@spinframework/build-tools",
-      "version": "1.0.4",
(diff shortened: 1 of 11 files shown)


tschneidereit (approved):
Can we add a test for this by any chance? Other than that, LGTM!

karthik2804:
@tschneidereit I added a test but now realize it is going to fail until we have a new release as I added it to the template test . Do we want to do that or just add a new test that points at the latest build of the tools?

The more I think the more I feel like I should do the later as well as testing the templates can be run successfully.

tschneidereit:
> The more I think the more I feel like I should do the later as well as testing the templates can be run successfully.

Yeah, I think that's what we should do: in general we should be able to test changes before releasing them, instead of having failing tests that can only be fixed with a release. If we're doing the latter, we always run the risk of having some other issue causing the test to fail, which is hidden until we do the release.

radu-matei (approved):

PR #398, opened 2026-02-10
(build-tools): resolve issue where build failed when precompile sourcemap did not exist

karthik2804:
This also updates the testing example to use the package from main. This is to enable adding it to the tests to make sure that it works.

### DIFF
--- a/packages/build-tools/src/index.ts
+++ b/packages/build-tools/src/index.ts
@@ -72,19 +72,25 @@ async function main() {
     console.log('Componentizing...');
 
     const source = await readFile(src, 'utf8');
-    let { content: precompiledSource, sourceMap: precompiledSourceMap } = precompile(source, src, true, 'precompiled-source.js') as { content: string; sourceMap: SourceMapInput };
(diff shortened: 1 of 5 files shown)


packages/build-tools/src/index.ts
```diff
@@ -73,18 +73,24 @@ async function main() {
 
     const source = await readFile(src, 'utf8');
     let { content: precompiledSource, sourceMap: precompiledSourceMap } = precompile(source, src, true, 'precompiled-source.js') as { content: string; sourceMap: SourceMapInput };
```
tschneidereit: It probably makes sense to also change the type here, which would I think have caught this.
```suggestion
    let { content: precompiledSource, sourceMap: precompiledSourceMap } = precompile(source, src, true, 'precompiled-source.js') as { content: string; sourceMap: SourceMapInput | null };
```

tschneidereit (approved):
Makes sense to me

packages/build-tools/src/index.ts
karthik2804: Yep! Will add it.

PR #428, opened 2026-04-07
bump mocha to pass npm audit

karthik2804:
This bumps `mocha` to use a beta release to pass `npm audit` while I work to move audits to be run using pnpm/yarn where we can bypass some of the warnings if they do not apply to us.


radu-matei:
Are using a beta version here because it addresses the vulnerability?

karthik2804:
@radu-matei It bumps the dependencies which were causing `npm audit` failures. There is no stable release yet because the vulnerabilities do not apply to Mocha. I just wanted to bump it to get CI passing until I update audit using pnpm or yarn where we can use exceptions.

radu-matei (approved):

PR #429, opened 2026-04-07
(package/build-tools): update wit tooling dependencies

karthik2804:
Update wit tooling deps and some minor cleanup. Adds some tests for the build caching as well.


radu-matei (approved):
Yay, more tests!

PR #435, opened 2026-04-20
Update componentizeJS dependency.

karthik2804:
This requires us to bump both the build-tools package and the http-trigger as ComponentizeJS works with specific versions of the the wasi:http trigger.

This commit also re-enables AoT compilation with weval and adds a test to verify building with it.

Note: this also removes the `wit` of the cli world from the build-tools. This was originally added to workaround a bug in StarlingMonkey around the time the debugger was added. This is no longer needed. This also means that the we can now move the `build-tools` package to the `devDependencies` section which I feel reflects its purpose better.


packages/http-trigger/wit/wasi-http@0.2.10.wit
tschneidereit, line 1: This file should be renamed to reflect the new version

test/aot-test/README.md
tschneidereit, line 1: Either update this to explain what the test does, or (probably preferably) simply delete it?

tschneidereit (approved):
Looks good with nits addressed

PR #436, opened 2026-04-21
enable using spin-dependencies.wit during builds

karthik2804:
Updates the build tool to take into consideration the `spin-dependencies.wit` which is generated since Spin 4.0 to enable using dependencies easily with the SDK.


packages/build-tools/src/index.ts
```diff
@@ -56,6 +57,18 @@ async function main() {
       });
     }
 
+    // Handle spin-dependencies.wit which will be present if component has dependencies defined in spin.toml
+    let spinComponentDeps = getSpinComponentDependencies();
+    if (spinComponentDeps) {
+      console.log('Found Spin component dependencies');
```
tschneidereit: Should we remove this? If we do want to include logging for things like this, we might want to fine-tune the message a bit, at least, but I think just removing it might make sense.

test/deps-test/README.md
tschneidereit, line 1: Same as for the last PR: either remove this readme, or change its contents to something meaningful instead of the template contents.

test/deps-test/fibonacci.wasm
tschneidereit, line 1: I'm not a fan of adding a binary wasm file here. Let's maybe use a JS dependency instead and compile it from source? Since there's no benchmarking here, it certainly doesn't need to be fibonacci :)

tschneidereit (changes requested):
This looks good, except for small nits plus the binary wasm file, which we should replace.

test/deps-test/fibonacci.wasm
karthik2804: Makes sense. Added a simple JS component that gets built as part of the test.

test/deps-test/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "deps-test",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build-dependency-component": "node build-depedency-component.mjs",
```
tschneidereit: With the file itself renamed:
```suggestion
    "build-dependency-component": "node build-dependency-component.mjs",
```

tschneidereit (approved):
LGTM!

PR #455, opened 2026-04-23
bump typescript package to v6.0.3

karthik2804:
Mostly mechanical changes but did run into one bug with defining the types for `del` command under redis which has now been fixed.


tschneidereit (approved):
Thank you!

PR #456, opened 2026-04-24
add build tools as a peer dependency to other packages

karthik2804:
Adds the peers peer dependencies on to all packages so that we can catch package incompatibilieis during npm install. Only the `http-trigger` is tightly coupled because of the versions of `wasi:http`.

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -5,7 +5,9 @@ cd ..
 for d in packages/*; do
     echo "Building $d"
     cd $d
-    npm install
+    # Legacy peer deps will allow us to ignore the unmet peer dependency warning for build-tools, which is not needed for building the packages themselves.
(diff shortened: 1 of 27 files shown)


test/test.sh
```diff
@@ -5,7 +5,9 @@ cd ..
 for d in packages/*; do
     echo "Building $d"
     cd $d
-    npm install
+    # Using legacy-peer-deps here is okay because, we do not need the `build-tools` package to build the packages.
+    # We only need it when we build apps that consume said packages.
```
tschneidereit: It'd be good to change this comment to say what `--legacy-peer-deps` does. I assume it means that a missing peer dep doesn't cause an error?

tschneidereit (approved):
Looks great!

PR #464, opened 2026-04-28
Bump package versions and deprecate spin-postgres3

karthik2804:
This PR bumps all the version of the package by a patch version so that we can do a new release with the proper peer dependency on `build-tools` specified. 

This second commit removes the spin-postgres3 package in favor of maintaining a singular package which supports only the latest version of the Spin postgres interface. Currently it is not the latest version because the runtime still does not support P3 but we can update it once that is ready.


packages/spin-postgres/src/index.ts
```diff
@@ -64,3 +237,46 @@ function createPostgresConnection(
...
+      sanitized.push({ tag: 'int64', val: k });
+      continue;
+    }
+    if (typeof k === 'number') {
+      isFloat(k)
+        ? sanitized.push({ tag: 'floating64', val: k })
+        : sanitized.push({ tag: 'int32', val: k });
+      continue;
+    }
+    if ((k as any) instanceof Uint8Array) {
```
tschneidereit: I'm surprised the `as any` cast is required here, given that `Uint8Array` is a variant of the enum 🤷🏻

tschneidereit (approved):
LGTM!

packages/spin-postgres/src/index.ts
karthik2804: Good catch. This revealed a bug. It was needed because of type exhaustion which revealed the bug. 

Since, Uint8Array is `typeof` object. We were checking that at the start and by the time we got all the way around in the if matching, the type of `k` was `never`. I similarly realized that the null check was not correct as well. 

I will  look into adding more tests for this particular package.

PR #465, opened 2026-04-29
Add release ci

karthik2804:
This PR sets up CI to publish packages to NPM on tags. Once this is merged, we should also enable OIDC in NPM.

This removes the existing release action and copies over the parts for publishing docs. We also enable provenance in our packages now.


.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,126 @@
...
+        - name: Install Rust toolchain 
+          run: rustup target add wasm32-wasip1 wasm32-unknown-unknown --toolchain ${{ env.RUST_VERSION }}
+
+        - name: Install Node.js
+          uses: actions/setup-node@v6
+          with:
+            node-version: ${{ env.NODE_VERSION }}   
+        - name: Install npm dependencies
+          run: | 
+            # Install dependecies for generating docs
```
tschneidereit:
```suggestion
            # Install dependencies for generating docs
```

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,126 @@
...
+            npm install
+            # Install dependecies for all the packages
```
tschneidereit:
```suggestion
            # Install dependencies for all the packages
```

tschneidereit (approved):
Excellent!

PR #469, opened 2026-04-29
update templates to use latest @spinframework packages

karthik2804:
`build-tools` can now be moved into the devDependency because we no longer need wasi:cli (the bug in the starlingmonkey debugger that made this needed has now been fixed).

### DIFF
--- a/templates/http-js/content/package.json
+++ b/templates/http-js/content/package.json
@@ -13,7 +13,8 @@
   "license": "ISC",
   "devDependencies": {
     "mkdirp": "^3.0.1",
-    "esbuild": "^0.25.8"
+    "esbuild": "^0.25.8",
+    "@spinframework/build-tools": "^2.0.0"
   },
   "dependencies": {
     {%- case http-router -%}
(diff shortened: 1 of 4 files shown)


templates/http-js/content/package.json
```diff
@@ -13,7 +13,8 @@
   "license": "ISC",
   "devDependencies": {
     "mkdirp": "^3.0.1",
-    "esbuild": "^0.25.8"
+    "esbuild": "^0.25.8",
+    "@spinframework/build-tools": "^2.0.0"
```
tschneidereit: I thought the build-tools package didn't need to be listed anymore at all? Wasn't that the entire goal with the whole peer-deps thing?

karthik2804: It does not need to be strictly specified anymore yes. I am happy to remove the explicit mention in package.json if we prefer that. I just have a very slight preference that it be specified  as that package is imported in the build script.

The goal with peer-deps was to make sure that we inform users if they are trying to use incompatible versions of the HTTP trigger package and build tools.

tschneidereit: While I don't disagree that importing from a package that's not mentioned anywhere isn't ideal, I think it still causes less friction than having to upgrade two dependencies instead of one when really only one of them has interesting surface. So in my view the goal of this change has always been to remove the dependency on built-tools entirely :)

Perhaps we could instead add a comment to the build script explaining that `build-tools` is brought in by all the other `@spinframework/*` packages and thus is always available?

karthik2804: That seems like a reasonable thing to do.

tschneidereit (approved):
Thank you!

PR #504, opened 2026-05-12
add dependency bindings generation npm script

karthik2804:
Actually setup the package.json to help with consuming Spin component dependencies in JS.

### DIFF
--- a/templates/http-js/content/package.json
+++ b/templates/http-js/content/package.json
@@ -4,8 +4,9 @@
   "description": "{{project-description}}",
   "main": "index.js",
   "scripts": {
-    "build": "node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
-    "build:debug": "node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build:debug": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "generate-dependency-types": "node -e \"if (!require('fs').existsSync('spin-dependencies.wit')) process.exit(1)\" && jco guest-types spin-dependencies.wit -n root || node -e \"process.exit(0)\"",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "keywords": [],
@@ -13,7 +14,8 @@
   "license": "ISC",
   "devDependencies": {
     "mkdirp": "^3.0.1",
-    "esbuild": "^0.25.8"
+    "esbuild": "^0.25.8",
+    "@bytecodealliance/jco": "^1.17.6"
   },
   "dependencies": {
     {%- case http-router -%}
--- a/templates/http-js/content/package.json
+++ b/templates/http-js/content/package.json
@@ -4,9 +4,8 @@
   "description": "{{project-description}}",
   "main": "index.js",
   "scripts": {
-    "build": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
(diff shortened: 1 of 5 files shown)


templates/http-js/content/package.json
```diff
@@ -4,16 +4,18 @@
   "description": "{{project-description}}",
   "main": "index.js",
   "scripts": {
-    "build": "node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
-    "build:debug": "node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build:debug": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "generate-dependency-types": "node -e \"if (!require('fs').existsSync('spin-dependencies.wit')) process.exit(1)\" && jco guest-types spin-dependencies.wit -n root || node -e \"process.exit(0)\"",
```
tschneidereit: Isn't the idea to silently ignore a missing `spin-dependencies.wit` file and treat that as no dependencies existing? In that case, we should use `exit(0)` instead of `exit(1)`.

templates/http-ts/content/package.json
```diff
@@ -4,8 +4,9 @@
   "description": "{{project-description}}",
   "main": "index.js",
   "scripts": {
-    "build": "node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
-    "build:debug": "node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "build:debug": "npm run generate-dependency-types && node build.mjs && mkdirp dist && j2w -d -i build/bundle.js --initLocation http://{{project-name | kebab_case}}.localhost -o dist/{{ project-name | kebab_case }}.wasm",
+    "generate-dependency-types": "node -e \"if (!require('fs').existsSync('spin-dependencies.wit')) process.exit(1)\" && jco guest-types spin-dependencies.wit -n root || node -e \"process.exit(0)\"",
```
tschneidereit: Same as for the other template

tschneidereit (approved):
See the comments inline. It'd also be good to have tests for these things, but I appreciate that adding the right infrastructure for that is a bit of a big ask for this particular PR.

Similarly, I don't particularly like how massively complex the scripts have gotten, and would much prefer that complexity to live inside a single script that lives in the `build-tools` package. A key reason is that the current setup means we can basically never evolve the build tooling API without developers being forced to go through a painful manual migration step. Ideally, the templates' script section would be
```json
{
  "build": "build-component.js",
  "build:debug": "build-component.js"
}
```

Where `build-component.js` would be a script in the workspace's `bin` dir and would use defaults for everything not otherwise configured, and use [npm_lifecycle_event](https://docs.npmjs.com/cli/v8/using-npm/scripts?v=true#current-lifecycle-event) to know whether to do a release or debug build. It could take options to override those defaults, of course.

But again, nothing blocking this PR, I just wanted to spell it out while it's top of mind.

karthik2804:
@tschneidereit What would your opinion be if we moved all the complexity into the build.mjs file we already have setup for esbuild. That way the npm scripts can just call a single script instead of having to move that into the build-tools package. I do not have a fully formed opinion yet, so I will spend sometime evaluating the options.

templates/http-js/content/package.json
karthik2804: Good catch!

tschneidereit:
Moving the complexity into `build.mjs` wouldn't really be much better, IMO: the most important problem with the current setup is that it means changes to the build process require manual changes to the project's contents, instead of being part of a dependencies update.

karthik2804:
The part that I am struggling to wrap my head around a little bit is with moving all the configuration steps needed is the bundling part. I am not sure if we should be opinionated and take a dependency on Esbuild and if so how we would allow for the user to override configuration in case they want to use their own polyfills.  This is because we need the bundling step in between the generating bindings and componentizing.

Following up on 
> Where build-component.js would be a script in the workspace's bin dir and would use defaults for everything not otherwise configured

Even in this case, the use would have to have to manually edit the configuration options if the dependency update changes things similar to how it would be with API changes if they are non semver incompatible right?

edit: I think we maybe able to solve the first part by making the esbuild plugin do more than just add externals and take care of bindings generation and componentizing as well while keeping things backwards compatible.

tschneidereit (approved):
So much nicer, thank you!

PR #505, opened 2026-05-12
Update esbuil plugin

karthik2804:
Following up the discussion in #504 and an offline with @tschneidereit, this PR simplifies the build from the perspective of the user by making the esbuild plugin do the heavy lifting of generating types and componentizing. It also maintains backward compatibility of the plugin allowing the users to still use the `j2w` cli.


tschneidereit (approved):
Very nice!

PR #535, opened 2026-08-31
run tsc --no-emit on build to actually verify TS types

karthik2804:
Esbuild just does type stripping, it doesn't actually verify TypeScript types.


itowlson:
@karthik2804 Did this merge without a review?  That sounds like something we should be wary of - stuff going into main should be reviewed even if it's only pro forma.  If we don't have enough reviewers on this repo, or reviewers aren't responsive, then let's flag that.

Or am I misreading the history?

karthik2804:
@itowlson You are correct in that I should have waited for a review. I relied a bit too much on CI for the template change but I will make sure to get an explicit review next time.

itowlson:
Oh I was more thinking should we have _waves hands vaguely_ protection thingies set on this repo?

Issue #552, opened 2026-09-10
Spin JS/TS apps broken with `hono` upstream release of `4.12.9`


thorstenhans:
With the release of hono `4.12.9` all JS/TS are broken because they changed behavior of `fire`. See https://github.com/honojs/hono/issues/4632

IMO that should have been a major release.... but well... here we are 😄 

Intermediate fix for those users that use `hono` in a version newer than `4.12.8` is restoring the default behavior:

```ts
import { Hono } from 'hono';
import { fire } from 'hono/service-worker';

const app = new Hono();
//...

fire(app, { fetch: undefined });
```

itowlson:
Ugh, that's rude!  @karthik2804 are you okay with Thorsten's propose fix?  I guess this would affect only the templates?

thorstenhans:
Spoke with @karthik2804 about this, I'll create a PR with ^^

PR #557, opened 2026-09-15
Hono Regression & Manifest housekeeping

thorstenhans:
With this PR adds the following changes:

- (`http-ts|http-js`): Bump `hono` to `4.13.8` and address a regression introduced with `hono@4.12.9`
- (`http-ts|http-js`): Set `fetch` to `undefined` when launching the application (fix for regression ^^)

Following changes were made to all manifest templates (`http-js`, `http-ts`, `redis-js`, `redis-ts`):
- Add `spin.toml` schema reference
- All app manifest templates now have `allowed_outbound_hosts`
- separate component table `[component]` and component-build table `[component.{name}.build]` by a newline
- build command is consistently specified as an array


thorstenhans:
This fixes #552

karthik2804 (approved):

itowlson:
CI is failing on `npm audit` and, for some reason, refuses to run tests if audit fails.  As far as I can tell, this PR does not make any changes to the npm configuration except for upgrading `hono` so I don't know why the tests are suddenly red.

I am going to force merge this, but for the future we should separate the audit step from the test step, so that tests can run whether or not audit passes.  (My strong preference, as ever, would be to get audit off the "block a PR that didn't change anything" path.)

PR #560, opened 2026-09-17
run npm audit independant of tests

### LINKED ISSUE #558: CI: split test and audit
I'm looking at some Dependabot PRs.  All of them are red because of audit failures addressed by other dependabot PRs.

I would like to be able to see "are the tests passing, has this broken anything" but I can't because `npm audit` fails and so the tests are never run.

We should:

* Separate test and audit so I can view the results of those steps separately
* Run audit AFTER test so that a failed audit doesn't stop the tests running

karthik2804:
fixes #558


.github/workflows/test.yaml
```diff
@@ -50,6 +50,22 @@ jobs:
           cd test
           ./test.sh
 
+  audit_deps:
```
itowlson, line 53: Should this be after `test_template`, to maximise the coverage we get before it audit-crashes?

itowlson (approved):
One question/suggestion but in any case better than what we have today - thanks!

.github/workflows/test.yaml
karthik2804: They all run in parallel anyway. They do not depend on each other.

